PR #69, opened 2022-10-19
feat: add js file view, so now can edit js file in obsidian

windily-cloud:
1. Create a codemirror view, so can view and edit .js file in obsidian vault.
2. In the main.ts, register this view.

I simply tried it, and the process is similar to writing snippets in settings, but with more code highlighting and support for vim mode, the experience is much better. Do you see any other problems or optimizations?

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -13,6 +13,7 @@ import { colorPairedBracketsPluginLowestPrec, highlightCursorBracketsPlugin } fr
 import { cursorTooltipBaseTheme, cursorTooltipField } from "./editor_extensions/inline_math_tooltip";
 import { editorCommands } from "./editor_commands";
 import { parse } from "json5";
+import { JsfileView, JSFILE_VIEW_TYPE } from "./ui/js_file_view";
 
 
 export default class LatexSuitePlugin extends Plugin {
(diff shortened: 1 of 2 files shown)


artisticat1:
Thanks for your contribution!

This seems like a nice feature to have. However, I don't think it's a good idea to register a view for *all* `.js` files, since that's outside the scope of this plugin (which should contain only LaTeX-related features).

Are you able to apply the custom .js view only to the snippets file defined in the user settings?

windily-cloud:
How about changing the extname of the file? For example latex-suite.snippet. since obsidian is registered based on the file extname, I can't think of any other good way at the moment.

windily-cloud:
What do you think about registering other file extensions? For example `xxx.snippet`. I can't think of a better way at the moment because the obsidian registration function is run based on file extensions.

windily-cloud:
After I tried it, I found that editing js files in obsidian is not really a good idea, because I found many problems with.

1. after registering the view, if you don't register the extension `js`, the console will report an error. It's not clear what the problem is.
2. even if I detach the view, in the normal md file it still reports an error in the console, the view is still working.
3. after obsidian upgrade, I can't find a good reference code for event detach.
4. Different plugins registering the same extension can cause conflicts.

My level is limited and I think this is a very bad PR. I think editing snippets in vscode is the best solution for me. Thanks for the reply, I'll probably come back to PR if I get these issues sorted out. Hope I can fix it.😣😣😣

PR #75, opened 2022-10-27
load snippets from folder

### LINKED ISSUE #72: [FR] load snippets from multiple files
Hey, I really like this plugin, to the point where I will probably be switching from my current neovim + [luasnip](https://github.com/L3MON4D3/LuaSnip) + obsidian workflow to just working in Obsidian for convenience and for live preview.

The recent addition of loading snippets from a file is awesome and made me wanna switch since I can now easily edit and sync my snippets. One thing that bothers me still is that I am constrained to just one snippet file. Could you please extend this to allow to add an arbitrary amount of files (or all files in a folder) that can be loaded as snippets?

My usecase for this would be to have a file for each subject I'm taking so I can quickly find and tweak specific snippets and phase out old snippets I no longer need because I won't need some very course specific snippets or because the current professor has a different notation.

B4rc1:
This closes #72.

I'm pretty happy with the Implementation, I'm not sure how well this feature is communicated to the user.
Let me know if/where you want me to clarify something.


artisticat1:
Hey, thanks for your contribution!

- I think we could communicate better to the user that this option "loads snippets from all _files within_ a folder", as opposed to "loads snippets from a folder".
- Could you modify variables to use camelCase instead of snake_case?

B4rc1:
Do you want me to squash all these commits into one or should i leave them as multiple commits?

artisticat1:
Leaving them as multiple commits is fine!

Would you mind updating the setting description to address this?
> I think we could communicate better to the user that this option "loads snippets from all files within a folder", as opposed to "loads snippets from a folder".

B4rc1:
done

B4rc1:
This is not yet ready, priority does not work correctly, I will investigate in the coming days. Also automatic reload does not work.

B4rc1:
Okay, I implemented reloading of the snippets.

For the priority I'm stuck tho. I'm testing it like this: I have two files in a `Snippet` folder from which I load the snippets:

`file1.js`:
```js
[
    {trigger: "a", replacement: "b", options: "tAw"},
]
```
`file2.js`:
```js
[
    {trigger: "a", replacement: "c", options: "tAw", priority: 1},
]
```

to my understanding, the `c` replacement should then be used because of a higher priority value. Currently tho, its like the snippets are not being sorted at all because the `this.snippets` array looks exactly the same before and after sorting.

I am calling `this.sortSnippets(this.snippets);` tho and the code for `sortSnippets` looks fine aswell.

This does not work even if these snippets are in the same file, but loading from folder, but does work if loading the file directly. I am unfortunately at my wits end here. Could you please look into this?

artisticat1:
Sure! I'll try to have a look in the next few days.

artisticat1:
Merged! This will be available in the next release.

PR #90, opened 2022-11-25
fix #89

### LINKED ISSUE #89: [BUG?] short snippet has higher priority causing long snippet fail
plugin version: Latex Suite Ver 1.6.6

```javascript
...
    {trigger: "im", replacement: "\\mathrm{Im}", options: "mA"},
    {trigger: "lim", replacement: "\\lim_{ ${0:n} \\to ${1:\\infty} } $2", options: "mA"},
...
```
I found that the `'lim'` snippet will never be triggered since the `'im'` has higher priority.
image
image

I think snippets should be sorted by trigger length first before they're sorted by priority in the function `sortSnippet`  in this code file
[snippet_helper_function.ts](https://github.com/artisticat1/obsidian-latex-suite/blob/main/src/snippets/snippet_helper_functions.ts)

or whether I didn't config the plugin right?

YalienY:
longer snippet with another short snippet as its suffix won't fail now.

### DIFF
--- a/src/snippets/snippet_helper_functions.ts
+++ b/src/snippets/snippet_helper_functions.ts
@@ -9,6 +9,22 @@ import { parse } from "json5";
 
 
 export function sortSnippets(snippets:Snippet[]) {
+    // Sort snippets by trigger length so longer snippets will have higher priority
+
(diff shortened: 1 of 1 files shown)


artisticat1:
Good idea! Thanks for your contribution.

PR #95, opened 2022-12-07
fix: snippets not loading from folder correctly

B4rc1:
In the current version `this.getSnippetsWithinFolder` does not exist, because you refactored it into a different file with no class this function belongs to.

After removing the `this` everything works as expected


artisticat1:
Thank you for fixing this!

PR #99, opened 2022-12-12
fix mac command+Z undo sometimes not work as expected

### LINKED ISSUE #93: [Bug Report] `to` will auto expand to `\to` and cannot be undo with CMD+Z in Mac OS
Input `to` in math block will auto expand to `\to` and cannot be undo with CMD+Z in Mac OS (will input z when cmd+z), Wihle other expand operation can be undo correctly(/ expand to \frac{}{} can be undo).  I also search the config text and didn't find trigger of to.
Screen-2022-12-07-171749(1)

I have test it in obsidian sand box vault.

- Obsidian 1.1
- MacOS 13.0.1 (22A400) M1 MacBook pro
- Latex-suite 1.6.7

Yaozhuwa:
Fix #93 
In Macos, command+z not always work for latex suite snippets expansion.

For example, under default snippets, input `pi` will auto expand to `\pi`. However, command+z won't undo the expansion. It add a `z` after cursor instead. 

In function `handleKeydown = (key: string, shiftKey: boolean, ctrlKey: boolean, view: EditorView)`, you use ctrlKey to Allows Ctrl + z for undo, instead of triggering a snippet ending with z. In mac, it's `command+z` instead of `ctrl+z` trigger undo operation. And the command key corresponds to KeyboardEvent.metaKey.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "obsidian-latex-suite",
-	"version": "1.5.2",
+	"version": "1.6.8",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "obsidian-latex-suite",
-			"version": "1.5.2",
+			"version": "1.6.8",
 			"license": "MIT",
 			"devDependencies": {
 				"@codemirror/autocomplete": "^6.0.0",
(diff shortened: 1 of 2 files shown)


artisticat1:
Thank you for fixing this! 💚

PR #109, opened 2023-01-13
Add a literal \t to the default word delimiters

### LINKED ISSUE #108: The tab character might be useful as default as word delimiter, too
With this snippet defined:

```js
    {trigger: "=>", replacement: "$\\implies$", options: "tAw"},
```

When using it in normal text, it is expanded. But in the case of the start of an indented environment, such as a list point further explained, it is not.

Example:

```markdown
- this is the list item heading

	further explaining details
	=> conclusion
```

For now I've solved this in my local environment by typing tab in Neovim, which resulted in `	`, and then coyping it into the `Word delimiters` field at the end of the plugin settings. But since this is likely a common situation, it feels worth to me having this as a default.

I'm running 1.6.10 on Linux.

MultisampledNight:
Closes #108.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -54,7 +54,7 @@ export const DEFAULT_SETTINGS: LatexSuiteSettings = {
     taboutEnabled: true,
     autoEnlargeBrackets: true,
     autoEnlargeBracketsTriggers: "sum, int, frac, prod",
-    wordDelimiters: "., +-\\n:\t;!?\\/{}[]()=~$"
+    wordDelimiters: "., +-\\n\t:;!?\\/{}[]()=~$"
 }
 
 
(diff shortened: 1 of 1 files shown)


src/settings.ts
```diff
@@ -511,4 +511,4 @@ function createSnippetsEditor(content: string, extensions: Extension[]) {
     });
 
     return view;
-}
\ No newline at end of file
+}
```
MultisampledNight, line 514: Oh whoops, this change was not intended. I guess Neovim saved the file indirectly and added a newline here. Is that okay?

artisticat1: That's fine! Would you mind moving the `\t` so that it's located just after `\\n`? I think it makes sense for `:` and `;` to be sat next to each other.

PR #111, opened 2023-01-17
Respect the capitalized letters in LaTeX

waterlens:


artisticat1:
While "LaTeX" is more correct, I think "Latex" appears more friendly to casual users, so I'm going to close this for now.

Issue #100, opened 2022-12-21
[Bug] $ inside math blocks triggers inline math preview


mzietlow:
## Describe the Bug

There are multiple ways of using $ inside math blocks. For example, one could write
``` markdown
$$
\text{this is legitimite: $\pi$}
$$
```

or

``` markdown
$$
1+1=\$
$$
```
Unfortunately, the dollar sign triggers the inline math preview when the cursor is placed _before_ the last additional dollar sign:

### Cursor placement before the first additional dollar sign (issue appears)
![Cursor placement before the first additional dollar sign](https://user-images.githubusercontent.com/30702083/208873302-ffb27b0f-f9f3-4e64-b9be-f8d4b0567c20.png)

### Cursor placement between the additional dollar signs (issue appears)
![Cursor placement between the additional dollar signs](https://user-images.githubusercontent.com/30702083/208873447-d96f686a-ba70-4264-9e75-25cfeffbc60e.png)

### Cursor placement after the additional dollar signs (issue does not appear)
![Cursor placement after the additional dollar signs](https://user-images.githubusercontent.com/30702083/208873534-a3ecb917-d7b5-47ac-8d11-0ff0d91ab0c7.png)

## How to Reproduce

1. Open a new vault and enable Latex Suite with inline math preview.
2. Create a new file and paste one of my samples.
3. Place the cursor before the last additional dollar sign
4. A preview pop-up will appear

## Expected Behavior

No preview pop-up should appear, since we are inside a math block, not inside an inline math statement

artisticat1:
Fixed in 1.8.0.

Issue #49, opened 2022-09-06
Replacements with "$\d" or "[[\d"


Shesh6:
I've been trying to implement a text-mode regex snippet to convert any number to math mode after writing: e.g 3 -> $3$. I quickly discovered that basically any snippet which inserts a $ followed by a digit is bound to spaz out and crash the plugin.

I think this is due to the detection of the tabstop syntax. May I suggest adding an option to escape the $ character's tabstop functionality in the replacement syntax, seeing as it's a crucial symbol for working with LaTeX?

My current workaround is 3 -> $\hspace{0pt}3$ which works but it's messy.

A similar issue occurs with double brackets as far as I can tell. I use snippets to auto-prettify my Obsidian links and I'd run into problems if I was using numerical note titles or headers.

(P.S. Didn't want to add an issue for this, but I'm in love with this plugin (thank you), and I'm wondering if you considered a page/google sheet where members of the community can share our custom snippets!)

artisticat1:
Hello! Glad you're enjoying the plugin.


> I think this is due to the detection of the tabstop syntax. May I suggest adding an option to escape the $ character's tabstop functionality in the replacement syntax, seeing as it's a crucial symbol for working with LaTeX?

Yep, the problem is caused by the detection of tabstop syntax. I can add the ability to escape the $ character by inserting a \ before it.

> A similar issue occurs with double brackets as far as I can tell. I use snippets to auto-prettify my Obsidian links and I'd run into problems if I was using numerical note titles or headers.

Could you describe the problem with [[ brackets in more detail? It doesn't involve the $ character, so what happens there?

> I'm wondering if you considered a page/google sheet where members of the community can share our custom snippets!

Good idea! I just opened [this discussion page](https://github.com/artisticat1/obsidian-latex-suite/discussions/50) where people can share snippets with others, and I'll add a link to it in the readme.

Shesh6:
> Could you describe the problem with [[ brackets in more detail? It doesn't involve the $ character, so what happens there?

It's barely an issue on second thought, still I'll explain myself:

In the regex replacement syntax, [[\d]] is used to insert a captured group.

A clear example for this being an issue, if perhaps unlikely: if one were to use a note naming system that includes digits as note names, and would want to use regex to insert a markdown link to note "0", the replacement "[[0]]" would conflict with the captured group syntax. 

[...]

PR #132, opened 2023-03-24
feat: space trigger

StrangeGirlMurph:
Hey, don't be scared by the big number. I ran prettier on the project for better formatting. I added a smaller but pretty crucial thing I think. There now is a setting to select which key needs to be pressed for non-auto snippets to expand. The space bar is just way faster than tab and now everyone gets to simply decide on their own. I tested the code. I also formatted the readme + docs. I edited the $SYMBOLS and $SHORT_SYMBOLS.

Issue #130, opened 2023-03-14
[Feature Request] Preview Block Math in Source Mode


IntrovertHedgehog:
I really like the inline math preview function, and it works even in source mode. However, there's no math preview for block math in source mode (since it's default in preview mode I think). It would be nice to have a preview similar to that of inline math.

StrangeGirlMurph:
I am considering creating this. 

@artisticat1 was there a reason you didn't do it? 
I don't want to waste my time in case there is something weird/hard about this ^^

PR #133, opened 2023-03-29
feat: math preview

StrangeGirlMurph:
Added math preview for block equations. Also fixed https://github.com/artisticat1/obsidian-latex-suite/issues/100
The preview only gets displayed when the cursor is inside the equation though. Just didn't want/couldn't implement the preview when the cursor is between the $$ in a block equation. The algorithm for checking would just say that it's inside a inline equation. To make it consistent I just removed it. I think its absolutely no big deal.


StrangeGirlMurph:
Oh and of course https://github.com/artisticat1/obsidian-latex-suite/issues/130

StrangeGirlMurph:
Oh and I maybe also fixed https://github.com/artisticat1/obsidian-latex-suite/issues/49#issuecomment-1242495278 with being able to escape the $

StrangeGirlMurph:
Also yes this and the #134 pr both include the other commits from #132.
Wasn't my best idea... I am sorry. I worked on #132 on main and then started creating branches.
I can reverse this if you want. I haven't yet bc maybe you don't care. I also think this pr partly relies on #132.

PR #136, opened 2023-04-03
feat: add snippet setting

StrangeGirlMurph:
Adds an additional setting for the user to choose whether they want the snippet whitespaces at the end of an inline math block removed


artisticat1:
Thanks for your contribution!

mym0404:
Thank you Only unconfort was discarded!

PR #138, opened 2023-04-04
feat: math preview

### LINKED ISSUE #130: [Feature Request] Preview Block Math in Source Mode

StrangeGirlMurph:
Added math preview for block equations. This PR fixes #130, #100 and maybe https://github.com/artisticat1/obsidian-latex-suite/issues/49#issuecomment-1242495278

Now the preview only gets displayed when the cursor is inside an equation. Just didn't want/couldn't implement the preview when the cursor is between the $$ in a block equation. `isTouchingInlineEquation` would just say that it's inside an inline equation. To make it consistent I removed it for now and don't use the function at all. In my opinion it's really no big deal and can be dealed with at some other point if necessary. I didn't remove the function.


artisticat1:
It looks like this will cause the popup preview to be shown for block math in Live Preview mode. However, Live Preview already previews block math by displaying the rendered equation beneath the source. Therefore, the popup is unnecessary.

Can you modify this to only show the popup for block math in Source mode?

StrangeGirlMurph:
Oh yeah I forgot about that. Thanks! I don't really know how to implement that though. I don't know how to pass that information to the tool tip function. I don't know how codemirror and their StateFields work. Can you help me?

StrangeGirlMurph:
`this.app.workspace.getActiveViewOfType(MarkdownView)?.currentMode.sourceMode` should return:
- true on source mode
- false on live preview
- undefined on reading mode or if the current active view is not a markdown view
(It uses a bit of private API, however.)

artisticat1:
If there is a state field that stores whether the editor is in Source/LP mode (would have to investigate whether one exists), I think you could access it via `state.field(...)`. Then you could modify the boolean `shouldShowTooltip` based on this value.

StrangeGirlMurph:
Okay tell me when you found a way how to implement it or just go ahead and do it yourself :)

StrangeGirlMurph:
@artisticat1 any updates?

artisticat1:
At the moment, I don't have much time to look into it.

artisticat1:
It looks like you can use [this StateField](https://github.com/obsidianmd/obsidian-api/blob/583ba39e3f6c0546de5e5e8742256a60e2d78ebc/obsidian.d.ts#L976) to check whether or not the editor is in Live Preview mode!

StrangeGirlMurph:
Thanks! Should be good to go now :)

StrangeGirlMurph:
@artisticat1 just wanted to check in. It's been nearly two weeks since I added the fix. Have you had a chance to look at it? It's really a simple change.

artisticat1:
Apologies for the delay. Looks fine to me! Thanks for your contribution.

PR #140, opened 2023-04-04
feat: snippet trigger

StrangeGirlMurph:
I added the option to select which key needs to be pressed for non-auto snippets to expand. I also improved $SYMBOLS and $SHORT_SYMBOLS.


StrangeGirlMurph:
read https://github.com/artisticat1/obsidian-latex-suite/pull/141#issuecomment-1496575340

StrangeGirlMurph:
@artisticat1 :( 
Next time please look at the PRs and merge before refactoring the whole codebase......
With all the new files and everything someone basically has to completely rewrite the whole thing because you reorganized everything. Please keep this in mind for the future!

StrangeGirlMurph:
I did it with this one now but I am not sure I want to completely redo #138... :(

artisticat1:
> Next time please look at the PRs and merge before refactoring the whole codebase...... With all the new files and everything someone basically has to completely rewrite the whole thing because you reorganized everything.

Sorry, I should've given you a heads up! It was mostly just some functions and files that got moved around, though, so porting these changes should be straightforward.

Can you move any changes to the $SYMBOLS and $SHORT_SYMBOLS variables into a separate PR from this new space-trigger functionality?

StrangeGirlMurph:
Yes sure I can but is it really necessary?

artisticat1:
Apologies for the delay, I had other responsibilities! The changes for the new space-trigger feature look great and I'm happy to merge them.

However, the `${SHORT_SYMBOL}` variable is intended for short symbols that _should not_ automatically have `\` prepended to them (because it would likely trigger in unintended situations), but _should_ still automatically have spaces inserted after them when a letter is typed.

So "pm" and "mp" should be placed in `${SHORT_SYMBOL}`, not `${SYMBOL}`, and "sim" and "subset" don't really belong in `${SHORT_SYMBOL}`. "ln" is a frequently used operator that should be expanded to "\ln" by default, so doesn't belong in `${SHORT_SYMBOL}` either. (Although, now that I think about it, it would probably be best to create a new `${OPERATOR}` variable.)

Why has ";" been added to `${SYMBOL}`?

StrangeGirlMurph:
Yes so I am not sure if you've read the docs again. I also made some changes content wise to them. 
>`${SHORT_SYMBOL}` : Can be inserted in a `trigger`. Symbols which overlap with symbols from `${SYMBOL}`. Shorthand for the following: ... ([here](https://github.com/StrangeGirlMurph/obsidian-latex-suite/blob/4884b6550316feafa95f95867845708f97c9fb39/DOCS.md?plain=1#L81)) 

I followed your principle. SHORT_SYMBOL contains symbols that should be taken care of by themself because the characters in them collide with items of SYMBOL which would mean that with the auto flag they would trigger in unwanted situations.

pm and mp don't collide with any element of SYMBOL. That's why I put them there. I put ln in SHORT_SYMBOL because it collides with lnot. SYMBOL and SHORT_SYMBOL are not about what's being used often. It's about what needs extra care to prevent collisions.  For example the snippets could look like that to deal with them:
```js
  {trigger: "(${SYMBOL})", replacement: "\\[[0]] ", options: "rmA"},
  {trigger: "(^\\w)(${SHORT_SYMBOL})", replacement: "[[0]]\\[[1]] ", options: "rmA"},
```
This would prevent any unwanted behavior but is just one way to deal with them.

; is just as quad a symbol for spacing and should therefore be apart of it.

artisticat1:
I see, I hadn't really considered collisions/overlap between symbols. However, when considering what symbols to include in SYMBOLS, I think frequency of use should be taken into account. "ln" is used far more often than "lnot". So I would prioritise "ln" over "lnot", and have "ln" expand to "\ln" by default, instead of avoiding doing so to avoid a collision with "lnot".

I see that you've included some trig functions in SYMBOLS. If we're overhauling the SYMBOLS variables, I think it would be best to create an entirely new variable altogether for these, called OPERATORS.

> ; is just as quad a symbol for spacing and should therefore be apart of it.

`\;` is a spacing command, not a symbol, and so doesn't fit in with the rest of SYMBOLS. However, adding default snippets for some spacing commands (such as `\quad`) would be a nice idea.

If you move the symbol changes into a separate PR, I can merge the space-trigger feature now and we can discuss symbols there.

StrangeGirlMurph:
Fine I removed the symbol related changes and created a new PR see #145

PR #141, opened 2023-04-04
feat: prettify

StrangeGirlMurph:
I added the eslint and prettier script now separately. I recommend you merge, add an `.prettierrc` with you preferences and the run the script yourself and commit to main.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,10 +15,10 @@
     },
     "rules": {
       "no-unused-vars": "off",
-      "@typescript-eslint/no-unused-vars": ["error", { "args": "none" }],
-      "@typescript-eslint/ban-ts-comment": "off",
       "no-prototype-builtins": "off",
-      "@typescript-eslint/no-empty-function": "off",
-      "quotes": ["warn", "double"]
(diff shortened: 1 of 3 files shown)


StrangeGirlMurph:
@artisticat1 Now I made all the git mess go away and pulled the PRs all apart. I hope no code got lost during all that. I will check again tomorrow and run some more test on #140 #138.

artisticat1:
As I said in #132,

> I ran prettier, and found that in some areas it made code less readable by collapsing multiple empty lines into one, so I'd like to hold off on using it for now.

Unfortunately, there isn't a configurable rule for this in prettier. I'd be happy to try other code formatters though!

StrangeGirlMurph:
Okay well here is my take: I think this project would benefit a lot from a formatter. I think collapsing multiple empty lines into one is not a big deal and doesn't hurt the readability at all. If you want to visually separate code you can also do that with comments or just put the code into a function. I think weirdly placed empty lines only make it worse to be honest. It's your call. I'll just leave this PR open for now.

artisticat1:
I agree there is value in using a formatter and would be happy to consider other options.

PR #142, opened 2023-04-06
Prevent snippets in "\tag{}"

Anti-Apple4life:


src/main.ts
```diff
@@ -321,9 +321,15 @@ export default class LatexSuitePlugin extends Plugin {
...
-
+		if (withinEquation) {
+			if (isInsideEnvironment(view, pos, {openSymbol: "\\text{", closeSymbol: "}"}) || isInsideEnvironment(view, pos, {openSymbol: "\\tag{", closeSymbol: "}"})) {
+				withinMath = false;
+			} else {
+				withinMath = true;
+			}
+		} 
 
+
```
artisticat1, line 332:
```suggestion
		if (withinEquation) withinMath = !(isInsideEnvironment(view, pos, {openSymbol: "\\text{", closeSymbol: "}"}) || isInsideEnvironment(view, pos, {openSymbol: "\\tag{", closeSymbol: "}"}));
```

I think this is cleaner?

PR #145, opened 2023-04-21
feat: symbol variables

StrangeGirlMurph:
See https://github.com/artisticat1/obsidian-latex-suite/pull/140#issuecomment-1518039249

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -1,5 +1,7 @@
 # Documentation
+
 ## Snippets
+
 Snippets are formatted as follows:
 
 ```typescript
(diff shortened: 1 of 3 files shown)


StrangeGirlMurph:
Okay lets think about an approach to this. I guess letting the user set variables themself is necessary in some way.
The plugin could just have some defaults. IMO we could just have 2 Variables: GREEK and SYMBOLS containing everything. With a snippet like `{trigger: "([^\\d\\w\\\\])(${GREEK}|${SYMBOLS})", replacement: "[[0]]\\[[1]] ", options: "rmA"},` this wouldn't even cause collisions.

Issue #56, opened 2022-09-26
Add snippet options for inline and block math


fenjalien:
I want a way to change how a snippet behaves depending on if its triggered in an inline math block ($$) or a normal math block ($$$$). 
A notable usage would in snippets for fractions. When in a normal math block I want the `frac` function to be used, however when in an inline math block I want the `flatfrac` function to be used.
I would guess that this can be achieved by adding a new option symbol, something like `i` or `M`.

btw I've come from the quick latex plugin and I'm blown away by how well this works, great work :)

artisticat1:
Hello, glad you're liking the plugin :)

Sounds good, I can add two new snippet options for inline and block math.

artisticat1:
Implemented in 1.8.0. You can use `n` for inline math and `M` for block math.

Issue #30, opened 2022-06-15
add an option to ignore case of trigger


justl00kingfor:
would be great if there is an option to ignore case of trigger.

if there is a way to set flag in the trigger on regex mode would be a workaround.

thanks for this great plugin!

justl00kingfor:
maybe just add support to javascript syntax like "/pattern/flags" in the trigger would be a simple solution.

Off topic: consider to add some reference to "text expansion" features on plugin description to help people discover it.

Thanks for your attention!

superle3:
Is made possible by #219, should be closed

PR #146, opened 2023-04-23
Add codeblock to options

### LINKED ISSUE #115: [Feature request] Add `c` as snippet mode
I'm using this awesome plugin together with your [TikZJax](https://github.com/artisticat1/obsidian-tikzjax) plugin, and it's a blast so far! There's only the minor nit that often I'd like to type snippets that

- should _only_ run in codeblocks (for example, a `\begin{$0}$1\end{$0}` helper that should auto-expand)
- should _never_ run in codeblocks (for example, `->` expanding to `\longrightarrow` in text mode)

I don't know TypeScript, but I'd love to implement this, if that'd be advisable and acceptable! From looking a bit around in the source code, I think `withinMath` would need to be changed to an enum containing all possible modes:

https://github.com/artisticat1/obsidian-latex-suite/blob/efb2936982c926db68a7790bfeccac55815f3b37/src/main.ts#L323-L324

Alongside of some pattern matching instead of these lines:

https://github.com/artisticat1/obsidian-latex-suite/blob/efb2936982c926db68a7790bfeccac55815f3b37/src/main.ts#L488-L493

Additionally, I'm not sure if it would be worth supporting math mode stacked ontop of codeblock mode _inside_ of a codeblock, in the case of typing `tikz`/`latex` codeblocks? I often find myself typing `$\alpha$` in labels or the like where `mk@a` would have been more comfortable. How should the mode look like as data structure, then?

MultisampledNight:
Closes #115 with the suggested structure of a `Mode`. Written by a complete TypeScript and JavaScript novice, so I have a few questions:

1. Two new structures are introduced, `Mode` and `Options`, in `src/snippets/options.ts`.

    a. They are both basically bitfields. Is the current (very lengthy) implementation fine? Or could they be shortened in any way, like setting `false` as default while allowing an override? Also thought about just using an integer, but not sure if that'd make it shorter or easier to handle. Or maybe the shorthands could also be mapped to the field names, since in JS everything seems to be a hashmap...

    b. Should they be `interface`s instead?

2. #56 suggests `i` and `M` for **i**nline math and block **M**ath, which have been both implemented. This is done in a backwards-compatible way, so that `m` works as a shorthand for both. Should `m` be removed instead?

3. `Snippet` can't directly contain `Options`, since `Snippet` is the direct representation of the snippet file — and as such also contains the shorthands, not `Options`. So currently, on each run through each snippet, the shorthands are parsed on-the-fly into `Options`, but that feels, uh, redundant. Could this be cached without duplicating `Snippet`? Or am I worrying too much?

A draft for now, since the actual functionality of setting the `code` field of `Mode` when in a codeblock hasn't been implemented yet.


artisticat1:
Hello, this looks nice!

1. Actually, I'm no JS/TS expert myself either. So I'd need to give this some thought!
2. Keeping `m` for backwards compatibility is a good decision. I wonder if there's a letter other than `i` we could use for the inline math option, because it might be a good idea to reserve `i` for case-insensitivity (#30). (The use of `i` for case-insensitivity would be consistent with other tools e.g. sed, regex.) I can't really think of any good options though.
3. You can resolve this by simply parsing the Options at the stage we parse the contents of the snippet file, converting them to an array of Snippets! This happens here: https://github.com/artisticat1/obsidian-latex-suite/blob/7c636f8f0a9d42678fa99dca1d9d0771772fd549/src/snippets/snippet_helper_functions.ts#L47-L68

src/snippets/snippets.ts
```diff
@@ -1,9 +1,23 @@
...
+
+export class ParsedSnippet extends RawSnippet {
+    options: Options;
+
+	constructor(raw: RawSnippet) {
+		super();
+		// this is slightly obscure but I did not find any better way to express
+		// "the parsed snippet should get everything the raw one has if not overridden"
+		Object.assign(this, raw);
+		this.options = this.options instanceof Options ? this.options : parseOptions(this.options);
```
artisticat1, line 19: You should be able to achieve this with the [spread operator](https://basarat.gitbook.io/typescript/future-javascript/spread-operator#object-spread), like
```typescript
{...raw, options: parseOptions(raw.options)}
```

src/snippets/snippets.ts
```diff
@@ -1,9 +1,23 @@
...
-    replacement: string,
-    options: string,
-    priority?: number,
-    description?: string
+import { Options, parseOptions } from "./options";
+
+export class RawSnippet {
+    trigger: string;
+    replacement: string;
+    options: string | Options;
```
artisticat1, line 6: Since `RawSnippet` is a direct representation of the snippet file, shouldn't `options` strictly be a string? (Or is the intent for users to be able to supply an `Options` object to specify options?)

MultisampledNight: I also thought that, but TypeScript won't let me override the type in the derived `ParsedSnippet` class without errors.

artisticat1: Ah I see. Since separating things into `RawSnippet`s and `ParsedSnippets` serves the idea that the representation of parsed snippets need not depend on that of raw snippets, maybe it'd be better to make `ParsedSnippet` its own independent class and not a derived one?

MultisampledNight:
(fwiw I'll pick up work on this again this weekend and rebase on top of main)

Slightly unfortunate that there's no performant solution that allows unlimited lookbehind/lookahead. But I guess that's okay.

src/snippets/snippets.ts
MultisampledNight: Implemented, thanks!

src/snippets/snippets.ts
MultisampledNight: Thanks!

MultisampledNight:
...interestingly even with your code, the codeblock detection is only triggered after the first two characters in each line, which means that snippets like `mk` and `dm` are still triggered inside codeblocks if they appear as the first two characters in any line, which would be especially nerving with shell scripts (see `mkdir`, which would expand to `$dir$` if typing by muscle memory).

Removing the `if (token.contains("codeblock")) {` improves this, triggering codeblock detection already after the first character, but this potentially implies a huge performance cost, given that we'd run the code at _every_ mode creation (which happens at every typed character) at most 100 times to bubble up parent nodes.

The choice on enabling code mode after 2 typed chars without additional cost, or enabling code mode already after 1 typed char but with potentially high cost is up to you. Maybe even a user-controllable option to control this could be worth considering.

MultisampledNight:
I also noticed that half of the codebase uses tabs as indent, and the other half 4 spaces, sometimes intermixed. Is that intended? Should I normalize it?

MultisampledNight:
Turns out Obsidian has by itself some pretty reliable API to list codeblocks in a file as @1C0D nicely communicated to me on Discord, which is what I've used for now.

Only issue with it is that it's done through the cache, which updates quite slowly — typing a bit subjectively faster already results in sometimes the code mode not being detected, and falling back to text mode instead. I suppose it could make sense to combine the approaches of accessing the syntax tree directly and the caching API, or maybe even some heuristics on top.

The heuristics could work based on:

- the last few keystrokes
- the last "true" code mode keystroke
    - currently it happens quite often that a line inside a codeblock is misdetected as the start of the codeblock, since between cache request and actual line request, things have moved (data races, yay)
- maybe even the kind of text written
    - text lines are generally rather long, and are usually lacking in punctuation, usually only featuring `?¿!¡:;,."` and those rather sparingly in comparison to text (actual characters to be bikeshed, should probably be configurable, localization will want to have a word)
    - code and math lines are rather short, usually featuring lots of punctuation like `.=:()[]{}_$#\/+-&|` and specific semi-consistent casing like `camelCase`, `snake_case` and `PascalCase`

At this point, I'm probably just dreaming and making problems where there are none. So I'll follow suit and actually dream now, lol.

MultisampledNight:
Oh and I've forgotten, there's still some machinery like autofrac that calls the editor helpers directly rather than going through mode, which remedies the comfortable usage of the `math` codeblock language. Maybe `Mode` needs to hold generic bound/content spans.

artisticat1:
> ...interestingly even with your code, the codeblock detection is only triggered after the first two characters in each line, which means that snippets like mk and dm are still triggered inside codeblocks if they appear as the first two characters in any line, which would be especially nerving with shell scripts (see mkdir, which would expand to $dir$ if typing by muscle memory).

Hmm, sounds odd. I do think the best/most pure approach would be parsing the `syntaxTree` directly from CodeMirror. Obsidian's metadata cache updates quite slowly, as you mention, so I'm not sure it's suited for providing real-time updates while typing. I can look into the issue when I get around to reviewing this, if you haven't solved it already!

> I also noticed that half of the codebase uses tabs as indent, and the other half 4 spaces, sometimes intermixed. Is that intended? Should I normalize it?

Whoops, that's definitely not intentional. Feel free to normalise to tabs.

MultisampledNight:
I messed with them a bit indeed, these were the results I've got by running them matrix-wise:

<details>
<summary>(collapsed since they're quite long, format is `pos, side`)</summary>
only results where the resolved token type name contained `codeblock` are listed


​## Before codeblock

​- Line before
- `pos + 1,  1`

putting something on the same line before the initial ```` ``` ```` seems to confuse obsidian, it won't render the codeblock then

​## In codeblock

​- After the initial ```` ``` ````, nothing after, with content in codeblock
	- `pos - 1, -1 | 0 | 1 | undefined`
	- `pos    , -1`
	- `pos + 1,  1`
​- After the initial ```` ``` ````, nothing after, without content in codeblock
	- `pos - 1, -1 | 0 | 1 | undefined`
	- `pos    , -1`
​- First two chars, nothing after
	- `pos - 1,  1`
	- `pos    , -1`
	- `pos + 1,  1`
​- Third char, nothing after
	- `pos - 1, -1 | 0 | 1 | undefined`
	- `pos    , -1`
	- `pos + 1,  1`
​- First char, content after it
	- `pos - 1, -1`
	- `pos    ,  1`
	- `pos + 1, -1 | 0 | 1 | undefined`
​- Second char, content after it
	- `pos - 1,  1`
	- `pos    , -1 | 0 | 1 | undefined`
	- `pos + 1, -1 | 0 | 1 | undefined`
​- Third char and any after, content after it
	- `pos - 1, -1 | 0 | 1 | undefined`
	- `pos    , -1 | 0 | 1 | undefined`
	- `pos + 1, -1 | 0 | 1 | undefined`
​- Before the final ```` ``` ````, with content before
	- `pos - 1, -1`
	- `pos    ,  1`
	- `pos + 1, -1 | 0 | 1 | undefined`
​- Before the final ```` ``` ````, without content before
	- `pos    ,  1`
	- `pos + 1, -1 | 0 | 1 | undefined`

​## After codeblock

​- After the final ```` ``` ````
	- `pos - 1, -1 | 0 | 1 | undefined`
	- `pos    , -1`
​- Next line
	- `pos - 1, -1`

</details>

Which looks to me like checking the tokens at

- `pos - 1,  1`
- `pos    , -1`
- `pos + 1,  1`

containing `codeblock` should hit most cases we want to have the code mode triggered in. Only issue with that is that it fails at multiple empty lines, since in those cases even `pos - 1,  1` and `pos + 1,  1` resolve to `Document`.

MultisampledNight:
Fwiw I solved the codeblock detection (well it can't detect if only 1 char has been typed in a line, but the only way I see to fix this with acceptable performance outside of codeblocks woudl be to check the cache instead). Not that it matters much though, I don't think there's any snippet that would want to be run in code mode and is only 1 character long. Would probably be worth documenting in README.md or the like.

For now I'm on my way of working through the existing functionality like autofrac and the like to make them use `ctx` instead of recomputing the environment state each time when called.

MultisampledNight:
(fwiw the conceal is broken atm)

Somehow I need to get the plugin settings into the CM plugins. Currently I'm doing this over a `StateField` that is managed alongside with the `EditorState`, but I think I'll need to work around this and use global variables. Whatever.

artisticat1:
It turns out the right way to configure CM extensions is [storing the settings in a `Facet`, and wrapping it in a `Compartment`](https://codemirror.net/examples/config/) (so that only one copy of the config is loaded per `EditorState`). [See here for an example from the Dynamic Highlights plugin](https://github.com/nothingislost/obsidian-dynamic-highlights/blob/master/src/highlighters/selection.ts).

However, I wrote this plugin when I was new to CodeMirror, so currently the plugin settings are accessed by passing `LatexSuitePlugin` directly to functions such as `runAutoFraction`, `autoEnlargeBrackets`, `runSnippets`, etc. This isn't ideal, since it creates a dependency on `LatexSuitePlugin` that isn't necessary. It'd be possible (and much neater) to write a CodeMirror extension that provides Latex Suite's features, but only accesses CodeMirror's API, and doesn't rely on the Obsidian API at all.

We can achieve that by removing the references to `LatexSuitePlugin` and instead storing/accessing the plugin settings in/through a `Facet` wrapped in a `Compartment`, as described above. I'd actually been meaning to refactor and do just that after this PR was merged (to avoid conflicts). Let me know if you want to give it a shot, or if you'd rather I did it (as originally intended)! And thank you for the work you've done on this already :)

MultisampledNight:
To be honest, I actually don't really want to work on this anymore. TypeScript/JavaScript causes so many headaches for me coming from Rust and Julia that most of the time I'm trying to figure out what type I'm dealing with here and how to work around some limitation that it becomes rather frustrating writing code with it.

So I'd appreciate it if I could take the offer, and you could work from the state I have pushed at the moment (I do have local changes which store the plugin in an `export let plugin`, and while that did work, it'd require much more effort than properly storing the settings I believe).

Only things left I originally wanted to do (but are not possible without having access to the settings inside CM plugins) were

1. making ```` ```math ```` codeblocks work just like real math blocks. Not the case atm since everything depends on `getEquationBounds` rather than calling `Context.getBounds`, which knows when to use codeblock and when to use math mode bounds)
2. respecting math blocks while in `latex` and `tikz` codeblocks, and switching to math mode out of code mode appropiately.
3. (actually possible already) making snippet exclusions configurable. The global `EXCLUSIONS` array in `src/snippets/{run_,}snippets.ts` seemed kind of suspicious to me, given that the logic behind it could be very much of use for custom snippets, too

But I suppose I'll leave that up to you. Or look at it again after you've rewritten the settings management. Thank you for your time either way!

artisticat1:
No worries. I am happy to pick up from where this leaves off. Re: the remaining things you wanted to implement:

1 and 2. I will look into implementing these - thank you for the pointer regarding `Context.getBounds`!
3. You're right - I used a global array here simply out of convenience (so I wouldn't have to implement exclusions for custom snippets as an entire feature). I haven't received any feature requests for it thus far, but it can certainly be implemented in the future.

artisticat1:
I finally got around to finishing this off recently and doing some refactoring. Here are some changes I made:

- Plugin settings are stored in a facet
- Changed the inline math option from "k" to "n" for i**n**line math, which I also think works well since "n" is similar to "m"
- Added partial support for `math` codeblocks being treated as math
  - The conceal and colorMatchingBrackets CM extensions don't work atm as they need to compute the value of a `Mode` at different positions in a document (other than the user's cursor), but currently each `Mode` attached to a `Context` corresponds only to the cursor's current position
- Removed "ignoreMathLanguages" for now, since detection of math mode isn't currently based on the location of `$` signs
- Gave `Mode` more fine-grained options (`codeMath` in addition to inlineMath and blockMath, `textEnv` for text environments within math)

It should be possible to implement bullet point 2 (resepecting math in `latex` and `tikz` codeblocks) now by looking at the locations of `$` signs in the block, if you'd be willing to take another look. Regardless, thank you for all the work you did on this!

PR #167, opened 2023-07-27
feat: add setting to change autofraction symbol

### LINKED ISSUE #98: [Feature request] Add the ability to adjust the triggering of the "x/y" command
Add the ability to adjust the triggering of the "x/y" command from `\frac{x}{y}` to `\dfrac{x}{y}`. I prefer to use `\dfrac` to write fractions. In the plugin settings in the snippet, it is possible to replace most conversions with` \dfrac`, but you cannot change the creation of fractions with "/" in this snippet.

adamhl8:
Closes #98 

Added a textArea setting that allows the user to change what symbol is inserted during autofraction replacement.

### DIFF
--- a/src/features/autofraction.ts
+++ b/src/features/autofraction.ts
@@ -104,8 +104,7 @@ export const runAutoFractionCursor = (view: EditorView, range: SelectionRange, p
         numerator = numerator.slice(1, -1);
     }
 
-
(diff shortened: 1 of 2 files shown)


artisticat1:
Looks good, thank you for the PR!

PR #195, opened 2023-09-18
include secant in trig functions

Comprehensive-Jason:
For some reason, secant isn't included in the default snippets for trig functions even though arcsecant is included. I added it in.


artisticat1:
Thanks!

PR #216, opened 2023-10-20
feat(conceal): conceal `\set{}` command

duanwilliam:
conceals usage of the `\set{}` command, replacing it with curly braces.

sample input:
```md
$$
x \in \set{ 0, 1 }^{*}
$$
```
unconcealed:
image
concealed:
image

---

i've mostly followed the coding style of the surrounding code, with a few deviations:
- use of a regex literal in favor of constructing regex from a string
- naming the regex literal `setRegex` as opposed to the generic `symbolRegex` used elsewhere
- space between `:` and `Concealment[]` in the two places where it's used. no space is how it's written from `concealFraction`, which i based this code off of, but there _is_ a space otherwise for parameter types
- single-line `if` statements are wrapped in braces regardless to make control flow more explicit

### DIFF
--- a/src/editor_extensions/conceal.ts
+++ b/src/editor_extensions/conceal.ts
@@ -346,6 +346,35 @@ function concealBraKet(eqn: string, selection: EditorSelection, eqnStartBound: n
 	return concealments;
 }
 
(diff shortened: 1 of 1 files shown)


artisticat1:
Looks good to me, thanks!

Issue #168, opened 2023-07-27
[DOCS]  how to format a file containing file snippets


userrand:
# Description

It is unclear to me how one is supposed to format a file containing snippets. What extension should the file have? 

Should there be more than just the snippets in the file ?

Another user mentioned their confusion regarding this here: https://github.com/artisticat1/obsidian-latex-suite/discussions/148
# Importance

Importance level: 4/10 (I can manage without it)

wenlzhang:
I also had the same question. After searing for the keyword "file", I found the following remark that answered my doubt:
https://github.com/artisticat1/obsidian-latex-suite/pull/75#issuecomment-1297370748

In short, one can following the following steps to achieve this:

- Copy and paste everything in the settings
- Put them in `.js` files and add own snippets
- Add the path to the folder where the `.js` files are stored to the settings
-

userrand:
> I also had the same question. After searing for the keyword "file", I found the following remark that answered my doubt: [#75 (comment)](https://github.com/artisticat1/obsidian-latex-suite/pull/75#issuecomment-1297370748)
> 
> In short, one can following the following steps to achieve this:
> 
> * Copy and paste everything in the settings
> * Put them in `.js` files and add own snippets
> * Add the path to the folder where the `.js` files are stored to the settings

Hello thanks, I tried following this but then none of the snippets worked and I got linting issues when i saved the file as a js file. Did you add anything else to the file ?

wenlzhang:
What kind of linting issues do you have? I did not add anything special, except for defining some snippets for my use case.

My `.js` file looks something like the following:

<img width="721" alt="image" src="https://github.com/artisticat1/obsidian-latex-suite/assets/44930227/43c7f07a-a12a-4b47-a2c4-b6b772d178fc">

userrand:
> 

Using solarlint vscode extension (which I am not very familiar with) i am getting the error below whereas the snippets work if i copy paste them to the settings text area

![image](https://github.com/artisticat1/obsidian-latex-suite/assets/114431098/f2478620-301d-4dfc-aba2-c33c4455dee8)

wenlzhang:
> Using the solarlint vscode extension (which I am not very familiar with) i am getting the error below

I also saw such errors when installing the extension. However, I did not do anything about them, and the plugin works in Obsidian.

mayurankv:
[...]

Issue #16, opened 2022-05-13
[feature request]: Support for function extension snippets


windily-cloud:
This plugin works well enough and very elegantly allows me to edit latex formulas.

If it could support function fragments, it would be perfect! My thoughts are as follows.

`mxi(3)` outputs a `3x3` unit matrix and `mxi(6)` outputs a `6x6` unit matrix.I can also customize `h(3)` to output three-level headings.

Also, it would be easier to debug and backup if the configuration file could be separated into an md file.

Thanks for any replies, I really love this plugin!

artisticat1:
> `mxi(3)` outputs a `3x3` unit matrix and `mxi(6)` outputs a `6x6` unit matrix.I can also customize `h(3)` to output three-level headings.

I'll need to put some thought into how to achieve this. I guess the goal is to allow executing code within a snippet, right?

> Also, it would be easier to debug and backup if the configuration file could be separated into an md file.

Good idea! I could add a button to export the user's snippets to a .txt or .md file, and a button to load snippets from a .txt / .md file. Would that work for you?

Your plugin settings (including snippets) should be backed up by git / any sync service you're using, btw.

> Thanks for any replies, I really love this plugin!

Glad you enjoy it!

windily-cloud:
> I'll need to put some thought into how to achieve this. I guess the goal is to allow executing code within a snippet, right?
I'm not quite sure there's a better way to implement this feature. My idea is to type the function name and parameters, then output the snippet returned by the function.

I've seen people implement it like this(I forget exactly where it came from, I retrieved it from last year's photo gallery😂):

![{9CF982E7-D6C5-FCEF-058A-18C449468135}](https://user-images.githubusercontent.com/30640638/168410211-c85df094-cad7-4bae-a894-40a642dfa255.jpg)

I guess his idea  is to use regular expressions to extract the function name and parameters, and then output the snippets returned by the function if it matches the preconfigured function.This avoids the case of entering more parentheses, and I don't know if there are any side effects.

![image](https://user-images.githubusercontent.com/30640638/168410570-8f3e03ef-2175-47c0-a3e6-da15fdfd24e9.png)

Of course, this is just an idea, we can have a more elegant approach. This method of his has to manually modify the code optimization himself when dealing with multiple parameters or when the parameters are arrays.

[...]

PR #219, opened 2023-10-21
feat: support javascript snippet files

duanwilliam:
(submitting this as a pr vs proposal issue because it seems there are a lot of issues backlogged, and makes tentative implementation easy to reference).

Proposal: add support for javascript snippet files.
doing so allows for "native" handling of variables (could largely overshadow the separate variables setting in obsidian) and more programmatic control over snippets.
for a trivial example, i don't like having to write out "trigger", "replacement", etc each time - so i could instead, say, define a function that lets me write shorthand `s({t: ..., r: ..., ...})`.
js also allows for directly writing regex literals, which allows for more concise, readable snippets (e.g. double-escaping backslashes, `{ trigger: "\\\\command", options: "r" }` -> `{ trigger: /\\command/ }`)

---
adds support for snippet files to be written as actual js files.
when parsing a snippet file, first tries to parse it as a js esm module, with the snippets array being the default export. that is, the following would be a valid snippets file:
```js
const SHORT_SYMBOL = "to|pm|mp"
export default [
  { trigger: `([^\\\\])(${SHORT_SYMBOL})`, replacement: "[[0]]\\[[1]]", options: "rmA" },
]
```
if that fails, falls back to existing parsing mechanism (json5). 
there's also some commented out code that could completely remove the need for json5 by prefixing it with `export default ` and parsing it as a module as well ([permalink](https://github.com/duanwilliam/obsidian-latex-suite/commit/6a2546fbc07c09af16082aa3c8b5e016a2c8d446#diff-eca73eac5a93841752e9144b7ac41451f7657632fce59e7984fba7a40ef2e1a5R57-R58)), since json5 is valid es5 js. only notable conflict i've noted between the two formats is that `${VISUAL}` can't really be replaced by a js variable and so needs to be written as `$\{VISUAL}` in js files (and, well, just generally, one can execute arbitrary code).

dynamically importing user-specified javascript modules notably does provide some security risk; i personally don't think it's too big of an issue, but it is something to keep in mind (and would mean users would need to vet shared snippets more carefully to ensure they are not doing something malicious).

---
additionally adds support for regex literals in js snippet files, and added a `flags` field that is applicable to any regex snippet, so e.g. the following is valid:
```js
[
  { trigger: "fOo", replacement: "bar", options: "rA", flags: "i" },
  { trigger: /BaZ/i, replacement: "qux", options: "A" },
]
```

also adds some brief docs explaining the snippets file format, since it seems several people have wondered about it (e.g. #148, #168).


artisticat1:
This is very cool! It also paves the way for allowing users to run code in snippet replacements by letting `replacement` be a `string | function`, as described in #16.

> only notable conflict i've noted between the two formats is that ${VISUAL} can't really be replaced by a js variable and so needs to be written as $\{VISUAL} in js files

I see. I don't think this is much of an issue as it's probably not very likely for ${VISUAL} and other template literals to be used in the same `replacement`. Then the `replacement` can simply be delimited by `"`s instead of backticks so it isn't treated as a template literal.

> there's also some commented out code that could completely remove the need for json5 by prefixing it with export default  and parsing it as a module as well ([permalink](https://github.com/duanwilliam/obsidian-latex-suite/commit/6a2546fbc07c09af16082aa3c8b5e016a2c8d446#diff-eca73eac5a93841752e9144b7ac41451f7657632fce59e7984fba7a40ef2e1a5R57-R58)), since json5 is valid es5 js.

The idea of prepending `export default` sounds good to me. This way users can use features like regex literals and (in future) function replacements without being concerned about JS modules. It also simplifies the appearance of the snippets file.

> dynamically importing user-specified javascript modules notably does provide some security risk; i personally don't think it's too big of an issue, but it is something to keep in mind (and would mean users would need to vet shared snippets more carefully to ensure they are not doing something malicious).

Perhaps we should add a note/warning in the docs regarding this.

duanwilliam:
> This is very cool! It also paves the way for allowing users to run code in snippet replacements by letting `replacement` be a `string | function`, as described in #16.
> 

ah yes that would be useful; i've made another branch on my fork exploring this (`replace-fn`), based off of this branch, that does work at the time of writing this. there are probably some design questions to explore there though, and this pr's overloaded enough. i put some notes/thoughts on it at https://gist.github.com/duanwilliam/f09624d29e62df23dc1c164e7e7da212

> > there's also some commented out code that could completely remove the need for json5 by prefixing it with export default and parsing it as a module as well ([permalink](https://github.com/duanwilliam/obsidian-latex-suite/commit/6a2546fbc07c09af16082aa3c8b5e016a2c8d446#diff-eca73eac5a93841752e9144b7ac41451f7657632fce59e7984fba7a40ef2e1a5R57-R58)), since json5 is valid es5 js.
> 
> The idea of prepending `export default` sounds good to me. This way users can use features like regex literals and (in future) function replacements without being concerned about JS modules. It also simplifies the appearance of the snippets file.
> 
> >

got it. i will note that the syntax highlighting for regex doesn't exactly play well with obsidian's default theme (see the gif earlier; dark blue on a dark background). seems to be set here:
https://github.com/artisticat1/obsidian-latex-suite/blob/b6a62f8f4294d92c6e1029ea3ff2c47204c4723b/src/settings/ui/snippets_editor/obsidian_theme.ts#L30

> > dynamically importing user-specified javascript modules notably does provide some security risk; i personally don't think it's too big of an issue, but it is something to keep in mind (and would mean users would need to vet shared snippets more carefully to ensure they are not doing something malicious).
> 
> Perhaps we should add a note/warning in the docs regarding this.

added a note in the readme on this branch, just underneath the sentence mentioning sharing snippets.

src/snippets/snippets.ts
```diff
@@ -1,22 +1,47 @@
...
+			// "d", // doesn't affect the search
+			// "g", // doesn't affect the pattern match and is almost certainly undesired behavior
+			"i",
+			"m",
+			"s",
+			"u",
+			"v",
+			// "y", // almost certainly undesired behavior
+		];
+		const resolvedFlags = Array.from(new Set(`${override.flags ?? ""}${raw.flags ?? ""}`.split("")))
```
artisticat1, line 40: Is it necessary to include `${raw.flags ?? ""}` when `override` already contains all the flags from `raw` (because `override` is defined as `{...raw}`)?

src/snippets/snippets.ts
```diff
@@ -1,22 +1,47 @@
...
 	options: Options;
+	flags: string;
 	priority?: number;
 	description?: string;
-
+	
 	constructor(raw: RawSnippet) {
-		const parsed = {...raw, options: Options.fromSource(raw.options)};
+		// normalize triggers as strings
+		const override: RawSnippet = { ...raw }
```
artisticat1, line 22: missing semicolon

src/snippets/snippets.ts
```diff
@@ -1,22 +1,47 @@
...
+			"s",
+			"u",
+			"v",
+			// "y", // almost certainly undesired behavior
+		];
+		const resolvedFlags = Array.from(new Set(`${override.flags ?? ""}${raw.flags ?? ""}`.split("")))
+			.filter(flag => validFlags.includes(flag))
+			.join(""); 
+
+		const parsed = {...raw, ...override, options: Options.fromSource(override.options), flags: resolvedFlags};
```
artisticat1, line 44: Same as above - I think just `...override` should be sufficient (instead of `...raw, ...override`)

artisticat1:
> ah yes that would be useful; i've made another branch on my fork exploring this (replace-fn), based off of this branch, that does work at the time of writing this. there are probably some design questions to explore there though, and this pr's overloaded enough. i put some notes/thoughts on it at https://gist.github.com/duanwilliam/f09624d29e62df23dc1c164e7e7da212

Ah, very nice! I'll have a think about the design and try to get back to you within a week or two.

> got it. i will note that the syntax highlighting for regex doesn't exactly play well with obsidian's default theme (see the gif earlier; dark blue on a dark background). seems to be set here:

Ah, I see. Replacing the color with `var(--text-accent)` should do.

> also, just realized template literals replacing variables settings might interfere with auto fraction (e.g. allowing spaces between greek letters for detection). i wonder if it might be better to move that into the default snippets (and possibly related utils like findMatchingBracket) at that point?

Hmm, I see how that could be a problem. However, I'm not sure I follow what you mean re: moving things into the default snippets? I agree it'd be nice to expose utils like `findMatchingBracket` to users though.

src/snippets/snippets.ts
duanwilliam: `override.flags` was clobbering that of `raw` in the case that trigger was a regexp. i've rewritten the related code in the constructor to make the values easier to trace and remove the use of `raw` outside of the initial construction of `override`.

duanwilliam:
> > also, just realized template literals replacing variables settings might interfere with auto fraction (e.g. allowing spaces between greek letters for detection). i wonder if it might be better to move that into the default snippets (and possibly related utils like findMatchingBracket) at that point?
> 
> Hmm, I see how that could be a problem. However, I'm not sure I follow what you mean re: moving things into the default snippets? I agree it'd be nice to expose utils like `findMatchingBracket` to users though.

hmm, guess i was thinking about the viability of having auto fraction settings deriving from snippets file as well instead of as another setting? something like
```js
const GREEK_SYMBOLS = ["alpha", "beta", /* ... */];
// some special export defining what can have a trailing space and still be captured in autofraction 
export const AUTOFRACTION_ALLOW_SPACES = [
  // this is kinda ugly, because it gets converted to regex it needs double backslashes
  ...GREEK_SYMBOLS.map(x => "\\\\x"),
  "foo",
  "bar",
  /* ... */
].join("|");
```
i do realize this is a slippery slope to go down though as it could make configuration much harder for those without programming backgrounds.

PR #220, opened 2023-10-25
fix(autofraction): only remove outer parentheses when they are matching

### LINKED ISSUE #206: [BUG] Auto-fraction strips parentheses from product of parenthesized terms
# Description

Making a fraction with auto-fraction and a parenthesized product like `(a)(b)` in the numerator removes the `( )` on the outside.
# To Reproduce

In a math environment, type `(n+1)(n+2)/`. The auto-fraction produces `\frac{n+1)(n+2}{}`.
# Expected Behavior

It should not remove the parentheses from the outside, leaving `\frac{(n+1)(n+2)}{}`. 

One possible solution would be to check if the regex described here https://stackoverflow.com/questions/546433/regular-expression-to-match-balanced-parentheses has more than one match (indicating that the numerator is a product and not just one term in parentheses). It seems that Obsidian regexes use the JS implementation though, so it might not be possible without scanning the entire string to count `(` and `)`.

duanwilliam:
resolves #206

output

### DIFF
--- a/src/features/autofraction.ts
+++ b/src/features/autofraction.ts
@@ -94,17 +94,22 @@ export const runAutoFractionCursor = (view: EditorView, ctx: Context, range: Sel
 		}
 	}
 
-	// Run autofraction
-	let numerator = view.state.sliceDoc(start, to);
-
 	// Don't run on an empty line
-	if (numerator === "") return false;
+	if (start === to) { return false; }
 
+	// Remove unnecessary outer parentheses
+	const { beg, end } = (() => {
+		if (curLine.charAt(start) === "(" && curLine.charAt(to - 1) === ")") {
+			const closing = findMatchingBracket(curLine, start, "(", ")", false, to);
+			if (closing === to - 1) {
+				return { beg: start + 1, end: to - 1 };
+			}
+		}
+		return { beg: start, end: to };
+	})();
 
-	// Remove brackets
-	if (curLine.charAt(start) === "(" && curLine.charAt(to - 1) === ")") {
-		numerator = numerator.slice(1, -1);
-	}
+	// Run autofraction
+	const numerator = view.state.sliceDoc(beg, end);
 
 	const replacement = `${settings.autofractionSymbol}{${numerator}}{$0}$1`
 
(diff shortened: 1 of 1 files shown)


src/features/autofraction.ts
```diff
@@ -94,17 +94,22 @@ export const runAutoFractionCursor = (view: EditorView, ctx: Context, range: Sel
...
 	// Don't run on an empty line
-	if (numerator === "") return false;
-
+	if (start === to) { return false; }
+
+	// Remove unnecessary outer parentheses
+	const { beg, end } = (() => {
+		if (curLine.charAt(start) === "(" && curLine.charAt(to - 1) === ")") {
+			const closing = findMatchingBracket(curLine, start, "(", ")", false, to);
+			if (closing === to - 1) {
```
artisticat1, line 104: I think this would be cleaner if we kept the original code, and simply moved the `findMatchingBracket` check inside the initial `if (curLine.charAt(start) === "(" && curLine.charAt(to - 1) === ")")` check?

duanwilliam: at the moment there's an implicit dependency between `numerator`, `curLine`, `start`, `to`; as i have it written at the moment mutation is minimized (no `let`s) and less string copying involved, albeit at the expense of being somewhat ugly with the iife. but even if that's removed i think i'd still favor something like
```js
if (start === to) { return false; }

let numerator = view.state.sliceDoc(beg, end);
if (numerator.at(0) === "(" && numerator.at(-1) === ")") {
    numerator = numerator.slice(1, -1);
}
```
to reduce some of those implicit dependencies - namely, still using `start === to` as the empty check and using `numerator.at()` to check its wrapping parentheses, as opposed to using `curLine`, `start`, and `to`, and i think this code still fairly resembles the original.

artisticat1: Ah, yes; using `start === to` and `numerator.at()` is definitely an improvement. I'd prefer to let `numerator` be mutable though; imo it's cleaner and involves less nested code.

artisticat1:
Looks like the `findMatchingBracket` check was removed, so in its current state the PR doesn't actually resolve #206?

duanwilliam:
oops, didn't look too closely or test it. should be fixed now

PR #226, opened 2023-11-26
add "\lceil: \rceil" and "\lfloor: \rfloor" to auto enlarge brackets

m4rch3n1ng:
i also thought about adding a snippet similar to the `"norm"` snippet for both of them, but while `"floor"` as a trigger would work, `"ceil"` would clash with the chemistry `"ce"` snippet, so i didn't. if you want i can add them later.

### DIFF
--- a/src/features/auto_enlarge_brackets.ts
+++ b/src/features/auto_enlarge_brackets.ts
@@ -23,7 +23,7 @@ export const autoEnlargeBrackets = (view: EditorView) => {
 
 	for (let i = start; i < end; i++) {
 
-		const brackets:{[open: string]: string} = {"(": ")", "[": "]", "\\{": "\\}", "\\langle": "\\rangle", "\\lvert": "\\rvert"};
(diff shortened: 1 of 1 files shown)


artisticat1:
Thanks!

> i also thought about adding a snippet similar to the `"norm"` snippet for both of them, but while `"floor"` as a trigger would work, `"ceil"` would clash with the chemistry `"ce"` snippet, so i didn't. if you want i can add them later.

Good idea. Perhaps the trigger of the "ce" snippet could be changed? I'm not sure what we'd change it to though.

m4rch3n1ng:
> Good idea. Perhaps the trigger of the "ce" snippet could be changed? I'm not sure what we'd change it to though.

you could do something similar to the `"inn"` => `"\\in"` snippet with `"cee"` => `"\\ce ... "`

artisticat1:
Sounds good, I'd be happy to accept a PR adding default snippets for "floor", "ceil" and "ce".

PR #241, opened 2023-12-18
feat(snippets): support replacement functions

### LINKED ISSUE #16: [feature request]: Support for function extension snippets

duanwilliam:
fixes #16.
initially touched on this in #219.

open to feedback/suggestions.

adds support for replacement functions _(replacers)_, taking the input and being able to transform it with javascript for replacement.
documentation is added, but somewhat bare. not sure the best way to approach it while keeping the basics simple for end users.

## dependencies

- [valibot](https://github.com/fabian-hiller/valibot) added for runtime validation of raw snippets. it's fairly lightweight and imo made validation a lot cleaner.
- typescript upgraded to latest version. i believe valibot depends on some newer typescript features, but either way having a newer typescript shouldn't hurt

## architecture

a `Snippet` (renamed from `ParsedSnippet`) is now one of three _types_ of snippets: visual, regex, or string. visual snippets operate on selections; regex snippets use regex; string snippets are any other type of snippet. a visual snippet is distinguished by the `v` option, or its replacement being a string with the substring `${VISUAL}`. a regex snippet is distinguished by the `r` option, or its trigger being a `RegExp` literal.
snippet types are _distinct_: a snippet cannot be both regex and visual, and a snippet that appears to satisfy both is considered invalid (i think technically this is a breaking change, but in practice i don't think anyone has a case like this?).
working with parsed `Snippets`, then, is simply done by matching on the snippet's `type` and performing actions accordingly.

all snippet parsing logic is moved into `parse_snippets.ts` (previously, parsing the raw snippets was there, while converting them to parsed `Snippet`s was in `snippets.ts`).
`sortSnippets`, previously in `parse_snippets.ts`, is moved to its own file `sort.ts`. i also rewrote it so it clearly sorts by priority, followed by trigger length.
`snippets.ts` now contains just snippet-wide information that may be useful elsewhere in the plugin: the `Snippet` type definition, the `${VISUAL}` magic string, and utilities for working with snippet-like things. math environment and environment exclusion information is moved from `snippets.ts` to a separate file `environment.ts`.

with the consolidation of parsing, additional checks have also been added to provide more useful error messages with invalid snippets. checks are also added to make parsing invalid snippets more robust. 

## replacer behavior

replacers are supported on all three types of snippets. replacers are expected to return a string; a snippet does not apply if a non-string value is returned (as such, if returning e.g. a number, it should be stringified).

_visual_ snippet replacers take a single parameter, the selection.
_regex_ snippet replacers take a single parameter, the `RegExpExecArray` result of `re.exec("string")`.
_string_ snippet replacers take a single parameter, the matched string.

## example

snippets

```js
[
  // string replacer
  { trigger: `hello`, replacement: hello => hello.split('').join(' '), options: 'A' },
  // visual replacer
  { trigger: "-", replacement: sel => { if (!sel.includes(" ")) { return false } return sel.replaceAll(/\s+/g, "-")}, options: "vA" },
  // regex replacer, #16
  { trigger: /mxi(?<n>\d)/,
    replacement: match => {
    const n = match?.groups?.n;
    let arr = [];
    for(let j = 0; j < n; j++) {
      arr[j] = [];
      for(let i = 0; i < n; i++) {
        arr[j][i] = (i === j) ? 1 : 0;
      }
    }
    let output = arr.map(el => el.join(" & ")).join(" \\\\\n");
    output = `\\begin{pmatrix}\n${output}\n\\end{pmatrix}`;
    return output;
    },
    options: 'mA',
  },
  {trigger: "dm", replacement: "$$\n$0\n$$", options: "tAw"},
]
```

usage

out


artisticat1:
Thank you very much for the PR! I'll try to review it as soon as I'm able.

DOCS.md
```diff
@@ -3,7 +3,14 @@
...
 ```typescript
-{trigger: string | RegExp, replacement: string, options: string, priority?: number, description?: string, flags?: string}
+{
+  trigger: string | RegExp,
+  replacement: string | ((str: string) => string) | ((match: RegExpExecArray) => string) | ((selection: string) => (string | false)),
+  options: string,
+  priority?: number,
+  description?: string,
+  flags?: string,
+}
```
artisticat1, line 13: Showing the full type definition for the `replacement` here will probably be overwhelming for users. My suggestion would be we leave this as `replacement: string`, and then create a new section lower in the docs titled "Function snippets", where we can show the full type definition, as well as include example snippets etc.
  
 Under the "Function snippets" heading I'd have something like
```md
Replacements can also be functions. **Function snippets** take the form {insert full type definition here}.
```
I'd be happy to help write the docs, by the way.

src/snippets/snippets.ts
```diff
@@ -1,58 +1,73 @@
...
+	trigger: string;
+	replacement: string | Fn<[match: RegExpExecArray], string>;
+	flags?: string;
+}
+
+interface StringSnippet extends _Snippet {
+	type: "string";
+	trigger: string;
+	replacement: string | Fn<[match: string], string>;
+	flags?: string;
```
artisticat1, line 55: Is this `flags` property for `StringSnippet`s ever accessed anywhere? (I'm pretty sure `flag`s are only used for regex snippets atm.)

I think it might actually make more sense to normalise the `trigger` for `RegexSnippet`s to be RegExps, instead of strings as we are currently doing. Then, we'd no longer need to define `const regex = new RegExp(`${trigger}$`, snippet.flags);` in `run_snippets.ts`, and we'd no longer need to store `flags` for regex snippets, since the flags would simply be contained inside the RegExp object.

From what I can tell, this would also simplify some of the parsing/normalisation logic used in `parse_snippets.ts`.

src/snippets/snippets.ts
```diff
@@ -1,58 +1,73 @@
...
+	function replacer(k: string, v: unknown) {
+		if (typeof v === "function") { return "[[Function]]"; }
+		if (v instanceof RegExp) { return `[[RegExp]]: /${v.source}/${v.flags}`; }
+		return v;
+	}
+	return JSON.stringify(snippet, replacer, 2);
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Fn<Args extends readonly any[], Ret> = (...args: Args) => Ret;
```
artisticat1, line 73: Not super familiar with TS -- is there a reason for defining this `Fn` type, as opposed to simply writing out the function signatures for each of the `replacement`s?

artisticat1:
Some initial thoughts (I haven't dived too deeply into the code yet):

- Classifying snippets into the three types (string/regex/visual) sounds good to me
- Never heard of valibot before, but improving type validation sounds great. Does this make it possible to show more precise error messages (e.g. errors pointing to a specific snippet causing issues) under the snippets editor to the user?

src/snippets/snippets.ts
duanwilliam: i think the major concern for this is it amounts to a pretty significant breaking change; but yes, flags are only ever relevant for regex snippets.

maybe there could be a codemod to transform string regex snippets, or even automatically do so for snippets defined in the editor, to try to mitigate the effects though, coupled with a deprecation period recommending switching to true regex if a string regex snippet is detected. but i'd still imagine it'd be quite a mess.

src/snippets/snippets.ts
duanwilliam: yeah it's not necessary, just there to make function signature types more concise.
would you prefer it be fully written out?

duanwilliam:
> Does this make it possible to show more precise error messages (e.g. errors pointing to a specific snippet causing issues) under the snippets editor to the user?

can't capture a specific line number, but yes we can print a serialized snippet (with replacement functions it'll just print those as `[[Function]]`, but at least trigger and options should be serialized properly); this is done by the `serializeSnippet` function in `snippets/snippets.ts` and doesn't need the validation library.

the validation library is more so of having a way to describe the expected shape of something and getting a validator + correct typescript types together from that single source of truth, v.s. defining a typescript type and manually creating validators to verify that values satisfy that type.
valibot honestly seems pretty new and its documentation isn't great, but i picked it for the time being for its small size. more featureful/mature alternatives [zod](https://github.com/colinhacks/zod) adds ~105kb to the plugin size, [typebox](https://github.com/sinclairzx81/typebox) ~95kb, while valibot adds only ~9kb (and the plugin as a whole, excluding them, is ~355kb)
in any case all of them should have errors that provide information that could be used to pinpoint which _part_ of a snippet is invalid as well, in addition to just which snippet errored as well.

DOCS.md
duanwilliam: added a section for `Function snippets` after snippet variables in `DOCS.md` and reverted type of `replacement` to `string` in the readme / snippet overview.

src/snippets/snippets.ts
artisticat1: Ah, I was referring to storing the trigger as a RegExp internally; users could still use string regex snippets perfectly well, the plugin would just normalise their string triggers to RegExps (as opposed to the other way round — normalising RegExp triggers to strings — which is what the code currently does)

(Currently, RegExp triggers are converted to strings. Then those strings are converted _back into_ RegExps when the snippet gets executed in `run_snippets.ts`. Makes sense to simply normalise to the "final form" the triggers are used in imo)

In future, I imagine we could also add flags for string snippets, such as "i" for case insensitivity (#30). To implement this we could simply internally map such string snippets to regex snippets with the same trigger, but with the "i" flag enabled.

src/snippets/snippets.ts
artisticat1: yes, thanks

artisticat1:
I see, thanks! Based on your explanation I'm happy to go with valibot for now as well.

> the `serializeSnippet` function

Would be nice to move this to a `toString` override method inside `Snippet`. Guessing this wouldn't really work though because `Snippet` is a type, not a class.

src/snippets/snippets.ts
duanwilliam: makes sense; but triggers still need to be strings in order to substitute snippet variables either way, and not doing so would probably amount to a major breaking change, seeing how the historical use case of the snippet variables is precisely for substituting in regex snippets as far as i can tell.

duanwilliam:
> Would be nice to move this to a `toString` override method inside `Snippet`. Guessing this wouldn't really work though because `Snippet` is a type, not a class.

if really needed, might be able to turn it back to a class, would need to double check but iirc something like this should work types-wise:
```
class Snippet<T extends SnippetType> {
  type: T
  data: SnippetData<T>
}
```

src/snippets/snippets.ts
artisticat1: Yeah, the intention would be to store the triggers as RegExps with snippet variables already substituted in (I have no intention of making a breaking change!).

If you're still opposed though, I suggest at least removing the unused `flag` property on `StringSnippet`, and moving the `trigger` property from the three derived snippet interfaces to `_Snippet` so we're only defining it once.

artisticat1:
I see. Perhaps turning it back into a class is something to consider?

Currently, `processSnippet` in `run_snippets.ts` works by matching on the snippet's type and performing actions accordingly, using a switch statement:

https://github.com/artisticat1/obsidian-latex-suite/blob/22728cb8ef2034bdfae5097799f8517677d6d5d4/src/features/run_snippets.ts#L116-L133

and, in all cases, returns a ProcessSnippetResult. What would your thoughts be on 

- Turning Snippet into an abstract class, moving `serializeString` to a method `toString` inside it, with `processSnippet` as an abstract method,
- creating three classes StringSnippet, RegexSnippet and VisualSnippet that each extend Snippet,
- moving the `processSnippet` logic for each SnippetType case into their respective classes,
- and replacing the `processSnippet` call in `run_snippets.ts` with a simple call to `processSnippet` on each Snippet instance?

This would remove the need to match on each snippet's type by delegating the implementation of the snippet logic to each derived Snippet class. I think `processSnippet` would be pretty much the only instance method though, so I'm not sure how much sense it'd make for Snippet to be a class.

duanwilliam:
personally tend to favor more functional-like approaches and don't really mind the switch statement, but seeing how the rest of the codebase is more imperative and oo-like it might be fair to move it into a class.

experimenting with refactoring it into a class at the moment, generally might not have the most immediate turnaround for the next few days though.

duanwilliam:
bulk of changes is listed in the commit, but summary:
- snippets refactored back into class, snippet types are subclasses
- regex snippets store trigger as _final_ regexp (including trailing `$`), computed at parse time
- exclusions are also computed at parse time now and stored as part of each snippet. i've also made it so each snippet stores exclusions as an array of excluded environments, so they can be excluded from multiple environments if needed without needing too much additional refactoring.
- `processSnippet()` logic is moved into `Snippet.process()`
- `sortSnippets` is refactored since triggers aren't always strings anymore. looks very different, but behavior should be the same, outside of the fact that it no longer mutates the array passed in so the the result needs to be assigned (and callers have been refactored accordingly)

artisticat1:
Ah, I hadn't considered that storing triggers as RegExps would create issues sorting by trigger length. Your solution looks good though. I did a small amount of refactoring as the normalised snippets now always take the form `{ type, trigger, replacement, options, priority, description }` and the normalisation step was adding a bit of unnecessary overhead, and fleshed out the docs. I'll push out a new update soon.

Thank you for your work!

Issue #255, opened 2024-02-12
[FEATURE] More than 9 tab stops


git-wil:
# Description of the Problem
Currently, there is no way to have snippets expand to have more than 10 unique tabstops, because the code to parse tabstops only parses a single digit after the dollar sign indicating the start of a tabstop. While this might not normally be very useful to many, with the introduction of function snippets, it is now easier than ever to create dynamically sized snippet outputs that can use many more than 10 tabstops, such as a large matrix.

# Description of the Solution
While I do not believe changing the default implementation of $ followed by a single digit for tabstops (because this would significantly break existing snippets), I propose to only modify the placeholder tabstop functionality to allow for tabstop numbers that are as long as necessary. Existing placeholder tabstops are parsed using the format `${X:YYY}` where X is a single digit and YYY is the replacement string. The current code for parsing these stops detects the prefix `${` and then looks for exactly one digit followed by a colon, and then parses the rest of the replacement string after. The only change that would need to be included is modifying the parser code to split the tabstop by the colon, and parse all text on the left as a multi-character number and all text on the right as the replacement string.

# Additional Comments
While there do already exist ways to create matrices with the matrix-shortcut feature, there exists some difficulty when using tab to add ampersands inside of a matrix and using tab to tabout of the current expression. That is, every time I try to add ampersands to my matrices using tab, I have to click tab several times to get rid of the backlog of tabstops that I skipped in order to actually produce the ampersand. I've already submitting another ticket for this issue, and while they are both important issues, this modified tabstop numbering feels easier to address more quickly and would be a useful feature for all.

I plan on submitting a pull request with suggested changes to the placeholder tabstop code soon!

PR #256, opened 2024-02-12
Modified tabstop replacement to allow for arbitrary length stop numbers in placeholder tabstops

git-wil:
This is an implementation of a recent feature request to allow for more than 9 tabstops in snippet replacements. I have modified the implementation of placeholder tabstops to allow for arbitrary length stop numbers by splitting the inside of the placeholder by the first colon. I have tested the modified `getTabstops` function locally and it seems to be behaving properly, but I expect further testing might be in order to make sure the modifications don't get in the way of efficiency or have any extreme edge cases.

### DIFF
--- a/src/snippets/codemirror/snippet_change_spec.ts
+++ b/src/snippets/codemirror/snippet_change_spec.ts
@@ -39,7 +39,7 @@ export class SnippetChangeSpec {
                 if (!(text.charAt(i+1) === "{")) continue;
     
                 // Find the index of the matching closing bracket
-                const closingIndex = text.indexOf("}", i)
+                const closingIndex = findMatchingBracket(text, i+1, "{", "}", false, start + this.insert.length);
                 
                 // Create a copy of the entire tabstop string from the document
                 const tabstopString = text.slice(i, closingIndex+1)
--- a/src/snippets/codemirror/snippet_change_spec.ts
+++ b/src/snippets/codemirror/snippet_change_spec.ts
@@ -42,13 +42,13 @@ export class SnippetChangeSpec {
                 const closingIndex = findMatchingBracket(text, i+1, "{", "}", false, start + this.insert.length);
                 
                 // Create a copy of the entire tabstop string from the document
-                const tabstopString = text.slice(i, closingIndex+1)
+                const tabstopString = text.slice(i, closingIndex+1);
     
                 // If there is not a colon in the tabstop string, it is incorrectly formatted
                 if (!tabstopString.includes(":")) continue;
     
                 // Get the first index of a colon, which we will use as our number/replacement split point
-                const colonIndex = tabstopString.indexOf(":")
+                const colonIndex = tabstopString.indexOf(":");
     
                 // Parse the number from the tabstop string, which is all characters after the {
                 // and before the colon index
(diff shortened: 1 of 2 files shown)


git-wil:
Fixes #255

src/snippets/codemirror/snippet_change_spec.ts
```diff
@@ -25,35 +25,48 @@ export class SnippetChangeSpec {
...
-                // Check for selection tabstops of the form ${0:XXX}
-                if (!(text.charAt(i+1) === "{" && text.charAt(i+3) === ":")) continue;
-
-                number = parseInt(text.charAt(i + 2));
+                // Check for selection tabstops of the form ${\d+:XXX} where \d+ is some number of
+                // digits and XXX is the replacement string, separated by a colon
+                if (!(text.charAt(i+1) === "{")) continue;
+    
+                // Find the index of the matching closing bracket
+                const closingIndex = text.indexOf("}", i)
```
artisticat1, line 42: You should use the `findMatchingBracket` utility function here because placeholder replacement strings can contain `{}`.

artisticat1:
Thanks for the PR. Will merge after I get around to testing this; do you happen to have any example snippets I can use to test?

git-wil:
Here is a snippet that you should be able to test with
```js
{trigger: "M([\\ pbvBV])(\\d)(\\d)", replacement: (m) => {
        var type = m[1] === ' ' ? '' : m[1]
        var n = 0
        var res = `\\begin{${type}matrix}`
        res += "\n"
        for (let i = 0; i < m[2]; i++) {
            for (let j = 0; j < m[3]; j++) {
                res += `\${${n++}:0}`
                if (j + 1 != m[3]) {
                    res += " & "
                } else {
                    if (i + 1 != m[2]) {
                        res += " \\\\"
                    }
                    res += "\n"
                }
            }
        }
        res += `\\end{${type}matrix}\${${n++}:}`
        return res
    }, options: "mr"},
```
This is a snippet to expand matrices. The replacement looks for a capital `M` followed by a single letter for the type of matrix (`b` for `bmatrix`, `p` for `pmatrix`, etc.) followed by a single digit for the number of rows, followed by a single digit for the number of columns. After pressing the trigger key, this should expand into a preformatted matrix filled with 0s that has tabstops for each entry and a final tabstop outside of matrix. The snippet works great so long as the number of elements is less than 9 (for example, `Mb33` creates a 3x3 `bmatrix` perfectly fine) but any more elements (for example, expanding `Mb44`) currently creates no tabstops after the first 10. The modified code for parsing these tabstops should however allow you to tab all the way through the matrix for as large as you'd like to make it.

artisticat1:
Thanks! I tried `Mp55` to create a 5x5 matrix, and the tabstops are created successfully, but when pressing `Tab` they get visited in the wrong order. More specifically, the tabstops are visited in the order `0 -> 1 -> 10 -> 11 -> 12 -> ...`. Can you reproduce this?

src/snippets/codemirror/snippet_change_spec.ts
```diff
@@ -25,35 +25,48 @@ export class SnippetChangeSpec {
...
+                // Check for selection tabstops of the form ${\d+:XXX} where \d+ is some number of
+                // digits and XXX is the replacement string, separated by a colon
+                if (!(text.charAt(i+1) === "{")) continue;
+    
+                // Find the index of the matching closing bracket
                 const closingIndex = findMatchingBracket(text, i+1, "{", "}", false, start + this.insert.length);
-
+                
+                // Create a copy of the entire tabstop string from the document
+                const tabstopString = text.slice(i, closingIndex+1)
```
artisticat1, line 45: There are a couple missing semicolons here (line 45, as well as lines 51 and 69).

artisticat1:
Thanks!

PR #257, opened 2024-02-19
Make the math tooltip appear below the line

### LINKED ISSUE #224: [FEATURE] Math Popup Preview below the line instead of above the line
# Description of the Problem
To be able to read the previous line that is above the line that I am currently editing with Latex, I either have to deactivate the Math Popup Preview or open a second tab with the same markdown file. This is very inconvenient if, for example, you are doing a proof and want/need to read the previous line.

# Description of the Solution
A simple solution would be to have the Math Popup Preview appear below the line that is currently being edited. I guess it is more likely that you will need what is above the line rather than what is below the line.

ps.:
Thanks for this great plugin, without it working with Latex in Obsidian would be a lot harder!

bdm-k:
Closes #224

### DIFF
--- a/src/editor_extensions/math_tooltip.ts
+++ b/src/editor_extensions/math_tooltip.ts
@@ -2,6 +2,7 @@ import { Tooltip, showTooltip, EditorView } from "@codemirror/view";
 import { StateField, EditorState } from "@codemirror/state";
 import { renderMath, finishRenderMath, editorLivePreviewField } from "obsidian";
 import { Context } from "src/utils/context";
+import { getLatexSuiteConfig } from "src/snippets/codemirror/config";
 
 export const cursorTooltipField = StateField.define<readonly Tooltip[]>({
 	create: getCursorTooltips,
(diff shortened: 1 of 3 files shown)


artisticat1:
Thank you for the PR. Can you make this a toggleable setting?

bdm-k:
Sure. However, since this is my first time developing an Obsidian plugin, it may take some time.

bdm-k:
https://github.com/artisticat1/obsidian-latex-suite/blob/27f3ac6c6575a8164c68de25de25cad60f37e4a7/src/settings/settings_tab.ts#L202-L203
`saveSettings` calls `refreshCMExtensions` internally, so line 202 is redundant. Is it all right if I correct this?

artisticat1:
I see, thanks for catching that. Yes, feel free to correct it!

artisticat1:
Thanks!

Aradoexis:
Thank you so much! ❤

PR #280, opened 2024-03-27
Feature: support snippet variable files.

itay-raveh:
# Feature: support snippet variable files.
## Changes

Introduces support for reading variables from files, almost exactly like the existing feature for snippets.

Since this isn't a small PR, I tried to change as little as possible. Specifically, I mostly made another variation of each snippet-loading function for variable-loading, rather than generalizing them. Happy to refactor.

## Motivation

I wanted to maintain separate folders, each with the setup for a specific workflow. This fits so well with the existing recursive snippet-file discovery, and now we'll have the same quality-of-life with variables.

Also, writing JSON in a plain `<textarea>` isn't fun, being able to use vscode/notepad++ is a nice bonus.

---

image
image


artisticat1:
Thank you for the PR! I'll try to review this in the next week.

I understand you've allowed snippet variables to be written in one of three different forms. However, I feel that sticking to `"${VARIABLE}": "..."` may be best for the sake of simplicity/clarity — it's easy to read this as a "find-and-replace" of `${VARIABLE}` with `...`. This is less true of `VARIABLE: "..."` and `"VARIABLE": "..."`. What do you think?

itay-raveh:
> Thank you for the PR! I'll try to review this in the next week.
> 
> I understand you've allowed snippet variables to be written in one of three different forms. However, I feel that sticking to `"${VARIABLE}": "..."` may be best for the sake of simplicity/clarity — it's easy to read this as a "find-and-replace" of `${VARIABLE}` with `...`. This is less true of `VARIABLE: "..."` and `"VARIABLE": "..."`. What do you think?

*For me*, defining the variables with `${}` in the name is less intuitive. It does not align with JavaScript, where a template string like ``​`${VARIABLE}`​`` refers to a variable named `VARIABLE`, not `${VARIABLE}`. The three forms are just a result of making the `${}` in the name optional, and the fact that JavaScript accepts both `{key: "..."}` and `{"key": "..."}` as valid and identical objects.

However, my intuition is entirely derived from preexisting JavaScript knowledge, which isn't an assumption we can make about all the plugin's users. Accordingly, I completely agree that the documentation should just stick to the single existing format, with the `${}`.

In contrast, we can still have the code that handles the case with no `${}`. That way, things will “just work” for users who share my intuition, or forget the `${}` by mistake.

DOCS.md
```diff
@@ -86,7 +86,11 @@ Using a RegExp literal, the same snippet can be written as
 
 
 ### Snippet variables
-Snippet variables are used as shortcuts when writing snippets. By default, the following variables are available for use in a `trigger`:
+Snippet variables are used as shortcuts when writing snippets. Strings with the format `"${VARIABLE}"` will be replaced with the contents of  `VARIABLE` if it's defined (and left unchanged otherwise).
+
+Snippet variables can be changed in the settings, under **Advanced editor settings > Snippet variables**. They are defined using a JavaScript object. For your convince `VARIABLE`, `"VARIABLE"` and `"${VARIABLE}"` are all accepted.
```
itay-raveh, line 91:
```suggestion
Snippet variables are used as shortcuts when writing snippets. In the `trigger` of a snippet, strings with the format `"${VARIABLE}"` will be replaced with the contents of the matching variable if it's defined, and left unchanged otherwise.

Snippet variables can be changed in the settings, under **Advanced editor settings > Snippet variables**. They are defined using a JavaScript object.
```

DOCS.md
```diff
@@ -224,16 +225,20 @@ based on which type of snippet the replacement applies to.
...
+Snippet/variable files can be saved with any extension. However, to obtain syntax highlighting in external editors, you may wish to save your snippet/variable files with an extension of `.js`.
 
 A snippets file is a JavaScript array of snippets, or a JavaScript module with a default export of an array of snippets.
 
 _Note: an **array** is a list of items. In JavaScript, this is represented with the bracket symbols `[` `]`, with the items between them and separated by commas, e.g. `[ item1, item2 ]`._
 
+A variables file is a JavaScript object, or a JavaScript module with a default export of an object.
+
+_Note: an **object** is a mapping of names to values. In JavaScript, this is represented with the braces symbols `{` `}`, with a `:` between the name and value, and a comma between name-value pairs, e.g. `{ name: "value", "name2": "value2" }`._
+
```
itay-raveh, line 241:
```suggestion
_Note: an **object** is a mapping of names to values. In JavaScript, this is represented with the braces symbols `{` `}`, with a `:` between the name and value, and a comma between name-value pairs, e.g. `{ "name1": "value1", "name2": "value2" }`._

```

DOCS.md
```diff
@@ -245,10 +250,22 @@ A snippet file containing an array of snippets:
...
 
+A variable file containing an object of variables, with the 3 legal formats:
+
+```typescript
+{
+  GREEK:
+    "alpha|beta|gamma|Gamma|delta|Delta|epsilon|varepsilon|zeta|eta|theta|Theta|iota|kappa|lambda|Lambda|mu|nu|omicron|xi|Xi|pi|Pi|rho|sigma|Sigma|tau|upsilon|Upsilon|varphi|phi|Phi|chi|psi|Psi|omega|Omega",
+  "SYMBOL":
+    "hbar|ell|nabla|infty|dots|leftrightarrow|mapsto|setminus|mid|bigcap|bigcup|cap|cup|land|lor|subseteq|subset|implies|impliedby|iff|exists|forall|equiv|square|neq|geq|leq|gg|ll|sim|simeq|approx|propto|cdot|oplus|otimes|times|star|perp|det|exp|ln|log|partial",
+  "${SHORT_SYMBOL}": "to|pm|mp",
```
itay-raveh, line 261:
```suggestion
  "${GREEK}":
    "alpha|beta|gamma|Gamma|delta|Delta|epsilon|varepsilon|zeta|eta|theta|Theta|iota|kappa|lambda|Lambda|mu|nu|omicron|xi|Xi|pi|Pi|rho|sigma|Sigma|tau|upsilon|Upsilon|varphi|phi|Phi|chi|psi|Psi|omega|Omega",
  "${SYMBOL}":
    "hbar|ell|nabla|infty|dots|leftrightarrow|mapsto|setminus|mid|bigcap|bigcup|cap|cup|land|lor|subseteq|subset|implies|impliedby|iff|exists|forall|equiv|square|neq|geq|leq|gg|ll|sim|simeq|approx|propto|cdot|oplus|otimes|times|star|perp|det|exp|ln|log|partial",
  "${SHORT_SYMBOL}": "to|pm|mp",
```

artisticat1:
> _For me_, defining the variables with `${}` in the name is less intuitive. It does not align with JavaScript, where a template string like `` ​`${VARIABLE}`​ `` refers to a variable named `VARIABLE`, not `${VARIABLE}`. The three forms are just a result of making the `${}` in the name optional, and the fact that JavaScript accepts both `{key: "..."}` and `{"key": "..."}` as valid and identical objects.
>  
> However, my intuition is entirely derived from preexisting JavaScript knowledge, which isn't an assumption we can make about all the plugin's users. Accordingly, I completely agree that the documentation should just stick to the single existing format, with the `${}`.
> 
> In contrast, we can still have the code that handles the case with no `${}`. That way, things will “just work” for users who share my intuition, or forget the `${}` by mistake.

That makes sense. I agree the documentation should prefer the existing `${VARIABLE}` format. However, if we omit the `VARIABLE: "..."` and `"VARIABLE": "..."` formats from the documentation entirely, how will users discover them?

I suggest sticking to the existing `${VARIABLE}` format in the _Snippet variables_ section, while outlining the other two formats under a new "Snippet variable files" section for more advanced users.

> Since this isn't a small PR, I tried to change as little as possible. Specifically, I mostly made another variation of each snippet-loading function for variable-loading, rather than generalizing them. Happy to refactor.

As for generalising the snippet-loading functions, yes, that would be preferable.

DOCS.md
```diff
@@ -224,16 +225,20 @@ based on which type of snippet the replacement applies to.
 
 If a snippet replacement function returns a non-string value, the snippet is ignored and will not expand.
 
-## Snippet files
+## Snippet/Variable files
```
artisticat1, line 228: I suggest splitting this out into its own "Snippet variable files" section.

src/snippets/parse.ts
```diff
@@ -143,11 +220,17 @@ function parseSnippet(raw: RawSnippet, snippetVariables: SnippetVariables): Snip
...
 
-		const normalised = { trigger, replacement, options, priority, description, excludedEnvironments };
+		const normalised = {
+			trigger,
+			replacement,
+			options,
+			priority,
+			description,
+			excludedEnvironments,
+		};
```
artisticat1, line 230: Could you undo this formatting change (and similar changes made elsewhere)?

itay-raveh:
> What's the intended behaviour when e.g. the snippet variables file is located inside the snippets folder? Or when the snippet variables folder is a subfolder of the snippets folder? I believe a common usage of the feature would include users creating a single folder as a "catch-all" for Latex Suite-related data, and storing their snippets and snippet variables there (and this is indeed what I tried to do myself). However, doing this currently raises an error.

Very good question, can't believe I didn't test this.

Below are some options we can implement to be able to determine which type of object is in a file. Tell me what you think.

### Automatically with `Array.isArray`

Just do:
```javascript
const imported = await importRaw(...);
if (Array.isArray(imported)) {
    // Parse variables
} else {
    // Parse snippets
}
```

This "just" fixes the bug, but it might be too "magical". 

### By requiring a file extension

Files that contain variables will have to be in the format `(optional-file-name.)variables.js`.

This allows, but also requires, explicit control by the user.

### By creating a unified format

Something like:
```javascript
export default {
    variables: {...}
    snippets: [...]
}
```
Where both keys are optional, and we aggregate all files and then parse both.

Probably overkill, just throwing it out there.

src/settings/file_watch.ts
```diff
@@ -20,100 +19,110 @@ function isInFolder(file: TFile, dir: TFolder) {
...
+		console.log(`Failed to load snippet/variable file ${file.path}:`, e);
 	}
-
-	return snippets;
 }
 
-async function getSnippetsWithinFolder(vault: Vault, folder: TFolder, snippetVariables: SnippetVariables) {
-	const snippets:Snippet[] = [];
-
+async function* getWithinFolder<T>(vault: Vault, folder: TFolder, parse: (content: string) => Promise<T>): AsyncGenerator<T> {
```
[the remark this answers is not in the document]
itay-raveh: Rather than return a value once, a generator `yield`s any number of values consecutively:
```javascript
function* foo() {
    yield 5;
    yield 1;
    yield 3;
}

for (const generated of foo()) {
    console.log(generated)
}

// 5
// 1
// 3
```

The `yield*` keyword makes a parent generator yield all values from a child generator:
```javascript
function* foo() {
    yield 5;
    yield 1;
    yield 3;
}

function* bar() {
    yield "start";
    yield* foo(); // read as `for (const generated of foo()) { yield generated; }`
    yield "end";
}

for (const generated of bar()) {
    console.log(generated)
}

// start
// 5
// 1
// 3
// end
```

In our case:

The old code was collecting the imported objects with `Object.assign` or `Array.concat` at each step of the recursion. This means each helper function must know which one to use.

Using generators, the values from all recursions are "streamed" into a top level loop which collects them. This means only the top level functions care if we're reading snippets or variables.

artisticat1:
> ### Automatically with `Array.isArray`
> Just do:
> 
> ```js
> const imported = await importRaw(...);
> if (Array.isArray(imported)) {
>     // Parse variables
> } else {
>     // Parse snippets
> }
> ```

> This "just" fixes the bug, but it might be too "magical".

Perhaps something along those lines, where we check whether each file

- lies inside the snippets folder only,
- lies inside the snippet variables folder only, or
- lies inside both.

We can then process each case accordingly:

- If a file lies inside the snippets folder only, process it as a snippets file.
- If a file lies inside the snippet variables folder only, process it as a snippet variables file.
- If a file lies inside both, attempt to parse it using your code snippet above.

I'm imagining something like this:

```typescript
const isSnippetsFile = settings.loadSnippetsFromFileOrFolder && isInFolder(file, snippetsFolder);
const isSnippetVariablesFile = settings.loadSnippetVariablesFromFileOrFolder && isInFolder(file, snippetVariablesFolder);

// Process each case
if (isSnippetsFile && isSnippetVariablesFile) {
    ...
}
else if (isSnippetsFile) {
    ...
}
else if (isSnippetVariablesFile) {
    ...
}
```

This also allows us to make user notices more specific and reduce ambiguity (e.g. have "Failed to load snippets file X" and "Failed to load snippet variables file X" as separate notices, instead of a single notice "Failed to load snippet/variable file X").

itay-raveh:
> I'm imagining something like this:
> 
> ```ts
> const isSnippetsFile = settings.loadSnippetsFromFileOrFolder && isInFolder(file, snippetsFolder);
> const isSnippetVariablesFile = settings.loadSnippetVariablesFromFileOrFolder && isInFolder(file, snippetVariablesFolder);
> 
> // Process each case
> if (isSnippetsFile && isSnippetVariablesFile) {
>     ...
> }
> else if (isSnippetsFile) {
>     ...
> }
> else if (isSnippetVariablesFile) {
>     ...
> }
> ```

Where would this occur? In the `onFileXXX` functions?

> This also allows us to make user notices more specific and reduce ambiguity (e.g. have "Failed to load snippets file X" and "Failed to load snippet variables file X" as separate notices, instead of a single notice "Failed to load snippet/variable file X").

Currently, error messages are raised from multiple places, including functions that are now generalized to handle both snippets and variables. I can try to refactor both the new and the existing code to bubble parsing errors rather than logging them. That is, until they reach a caller who knows if the error message should say "snippets" or "variables".

artisticat1:
> Where would this occur? In the `onFileXXX` functions?

Ah, I hadn't thought about where exactly this would occur – just what the overall logic would look like.

Here's an overview of a procedure I came up with. Let me know what you think.

1. Write a function `getFilesWithin(fileOrFolder: TAbstractFile): Set` that takes in a file or folder as input, collects all the `TFile`s within that folder, and returns them as a set.
    - (This would be similar to currently written `getWithinFolder` function.)
    - If `fileOrFolder` is a file, the set just contains that file.
3. Call `getFilesWithin` with the snippet file/folder location defined in the settings.
4. Call `getFilesWithin` with the snippet variables file/folder location defined in the settings.
5. Use [set composition functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#set_composition) to determine which files are snippet files, variable files, or snippet-or-variable files.
    - The `intersection` of the sets returned in steps 2 and 3 will give us the files that lie in both the snippet and snippet variable locations.
    - The `difference` will give us snippet-only files and variable-only files.
6. Parse variable files and snippet-or-variable files. If a snippet-or-variable file fails to parse correctly, add it to the snippet files set.
7. Parse snippet files.
 
> Currently, error messages are raised from multiple places, including functions that are now generalized to handle both snippets and variables. I can try to refactor both the new and the existing code to bubble parsing errors rather than logging them. That is, until they reach a caller who knows if the error message should say "snippets" or "variables".

I think the procedure I described above should solve this. (Except that at step 5/6, we may want to maintain a separate list of snippet-or-variable files that failed to parse as variable files, so if they also fail to parse as snippet files then we can print "Failed to parse snippet/variable file X")

src/settings/file_watch.ts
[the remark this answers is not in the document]
artisticat1: I see. Thank you for the explanation!

I'm personally in favour of changing the existing code from a recursive approach to a DFS/BFS-like approach, where we take in a `fileOrFolder: TAbstractFile` as input, traverse that folder, and return all the `TFiles` within it as an array/set. However, I'll leave it up to you if you feel that generator functions are the better way to go about things!

itay-raveh:
Generally followed this procedure, with a few minor teaks.

Set composition functions are [not available in Node.js](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#browser_compatibility), so I installed pollyfills from  [es-shims](https://github.com/es-shims)

Both success and error notices are now as specific as they can be.

src/main.ts
```diff
@@ -105,24 +97,54 @@ export default class LatexSuitePlugin extends Plugin {
...
+		 
+		new Notice(`Loaded ${Object.keys(snippetVariables).length} variables`)
+		
+
+		const snippets =
+			this.settings.loadSnippetsFromFile
+				? await getSnippetsFromFiles(this, files, snippetVariables)
+				: await this.getSettingsSnippets(snippetVariables);
+		
+		new Notice(`Loaded ${snippets.length} snippets`)
```
artisticat1, line 145:
```suggestion
		const snippets =
			this.settings.loadSnippetsFromFile
				? await getSnippetsFromFiles(this, files, snippetVariables)
				: await this.getSettingsSnippets(snippetVariables);
		
		new Notice(`Loaded ${snippets.length} snippets and ${Object.keys(snippetVariables).length} snippet variables.`);
```

I like these notices – however, currently the UI gets spammed with them when editing snippets or snippet variables from the settings page.

Can we get them to show only when
- the snippet/variable file/folder locations are modified in the settings, or
- a snippet/variable file is added/changed/deleted?

src/settings/file_watch.ts
```diff
@@ -20,100 +26,147 @@ function isInFolder(file: TFile, dir: TFolder) {
...
-		snippets = await parseSnippets(content, snippetVariables);
-	}
-	catch (e) {
-		new Notice(`Failed to load snippet file ${file.name}`);
-		console.log(`Failed to load snippet file ${file.path}:`, e);
-	}
+function getFilesWithin(vault: Vault, path: string): Set<TFile> {
+	const fileOrFolder = vault.getAbstractFileByPath(path);
+	const files = generateFilesWithin(fileOrFolder);
+	return new Set(files)
```
artisticat1, line 90: missing semicolon

src/settings/file_watch.ts
```diff
@@ -20,100 +26,147 @@ function isInFolder(file: TFile, dir: TFolder) {
...
+	const snippetsFolder =
+		plugin.settings.loadSnippetsFromFile
+		? getFilesWithin(plugin.app.vault, plugin.settings.snippetsFileLocation)
+			: new Set<TFile>();
+	
+	const definitelyVariableFiles = difference(variablesFolder, snippetsFolder);
+	const definitelySnippetFiles = difference(snippetsFolder, variablesFolder);
+	const snippetOrVariableFiles = intersection(variablesFolder, snippetsFolder);
+
+	return {definitelyVariableFiles, definitelySnippetFiles, snippetOrVariableFiles}
```
artisticat1, line 114: missing semicolon

src/settings/file_watch.ts
```diff
@@ -20,100 +26,147 @@ function isInFolder(file: TFile, dir: TFolder) {
...
 
-	for (const fileOrFolder of folder.children) {
-		if (fileOrFolder instanceof TFile) {
-			snippets.push(...await getSnippetsFromFile(vault, fileOrFolder, snippetVariables));
-		}
-		else if (fileOrFolder instanceof TFolder) {
-			const folderSnippets = await getSnippetsWithinFolder(vault, fileOrFolder, snippetVariables);
-			snippets.push(...folderSnippets);
+export async function getVariablesFromFiles(plugin: LatexSuitePlugin, files: FileSets) {
+	const snippetVariables: SnippetVariables = {}
```
artisticat1, line 118: missing semicolon

src/settings/file_watch.ts
```diff
@@ -20,100 +26,147 @@ function isInFolder(file: TFile, dir: TFolder) {
...
-	if (fileOrFolder instanceof TFolder) {
-		snippets = await getSnippetsWithinFolder(vault, fileOrFolder, snippetVariables);
-	}
-	else if (fileOrFolder instanceof TFile) {
-		snippets = await getSnippetsFromFile(vault, fileOrFolder, snippetVariables);
-	}
-	else {
-		return [];
+export async function tryGetVariablesFromUnknownFiles(plugin: LatexSuitePlugin, files: FileSets) {
+	const snippetVariables: SnippetVariables = {}
```
artisticat1, line 135: missing semicolon

src/settings/file_watch.ts
```diff
@@ -20,100 +26,147 @@ function isInFolder(file: TFile, dir: TFolder) {
...
+	if (!(file instanceof TFile)) return;
 
+	const snippetVariablesDir = plugin.app.vault.getAbstractFileByPath(plugin.settings.snippetVariablesFileLocation);
 	const snippetDir = plugin.app.vault.getAbstractFileByPath(plugin.settings.snippetsFileLocation);
-	const isFolder = snippetDir instanceof TFolder;
 
-	if (file instanceof TFile && (isFolder && file.path.contains(snippetDir.path))) {
-		refreshSnippetsFromFileOrFolder(plugin);
+	if (plugin.settings.loadSnippetVariablesFromFile && snippetVariablesDir instanceof TFolder && file.path.contains(snippetVariablesDir.path)
+	 	|| plugin.settings.loadSnippetsFromFile && snippetDir instanceof TFolder && file.path.contains(snippetDir.path)
```
artisticat1, line 72: there are mixed spaces and tabs on line 72

src/main.ts
```diff
@@ -105,24 +97,54 @@ export default class LatexSuitePlugin extends Plugin {
...
+				? await getVariablesFromFiles(this, files)
+				: await this.getSettingsSnippetVariables();
+		
+		// This must be done in either case, because it also updates the set of snippet files
+		const unknownFileVariables = await tryGetVariablesFromUnknownFiles(this, files);
+		if (this.settings.loadSnippetVariablesFromFile) { 
+			// But we only use the values if the user wants them
+			Object.assign(snippetVariables, unknownFileVariables);
 		}
+
```
artisticat1, line 136: same here (mixed spaces and tabs)

src/settings/file_watch.ts
```diff
@@ -1,9 +1,15 @@
...
 import { Vault, TFile, TFolder, Notice, debounce, TAbstractFile } from "obsidian";
 import { Snippet } from "../snippets/snippets";
-import { parseSnippets } from "../snippets/parse_snippets";
-import { sortSnippets } from "../snippets/sort";
-import type { SnippetVariables } from "src/snippets/snippet_variables";
+import { parseSnippets, parseSnippetVariables, type SnippetVariables } from "../snippets/parse";
+// @ts-ignore
+import differenceImplementation from 'set.prototype.difference';
+// @ts-ignore
+import intersectionImplementation from 'set.prototype.intersection';
```
artisticat1, line 8:
```suggestion
import differenceImplementation from "set.prototype.difference";
// @ts-ignore
import intersectionImplementation from "set.prototype.intersection";
```

artisticat1:
Works well from my testing! Looks good other than some minor formatting issues and the UI getting spammed with notices when editing snippets/variables from settings.

itay-raveh:
Fixed all the formatting issues. Consider writing a [.prettierrc](https://prettier.io/docs/en/configuration.html) file matching your style :)

As to the spam, I gave the notice a long debounce, and it's only created if at least one of them is loaded from a file.

artisticat1:
Thanks! Yeah, I've actually looked into writing a `.prettierrc` file before, but I wasn't completely happy with the output and some customisation options I was looking for didn't exist. Perhaps I'll look into another formatter in future!

itay-raveh:
> The upgrade to esbuild in [a04b339](https://github.com/artisticat1/obsidian-latex-suite/commit/a04b33920f90d4eefced78a1926b366e0e6d116c) causes the command `npm run build` to stall – any help? I'll revert the change temporarily so I can push out a release.
> 
> Edit: Nevermind, I appear to have solved the issue by switching between `esbuild.build` and `esbuild.context` depending on what the `production` flag is set to.

If you've solved it, then great, but I believe this upgrade can be reverted if you want.

I upgraded because the older version did not support async generators, but those are gone now. There is a single sync generator, which I believe the older version supports.

artisticat1:
Ah, that's good to know. I suppose we might as well leave it on the more recent version though.

PR #285, opened 2024-04-23
Backspace functionality to delete both $ in empty inline math

### LINKED ISSUE #266: [FEATURE] Automatically delete pair of `$` signs when backspacing in inline math
When pressing backspace on a line with `$|$` where `|` represents the cursor, both signs should delete, not just the first one.

Making this issue but I'll hopefully create a PR to fix this.

bdm-k:
Closes #266 

If the implementation looks good, I'd like to proceed with the next steps, such as:

* making this feature toggleable
* organizing the code into a new file

### DIFF
--- a/src/latex_suite.ts
+++ b/src/latex_suite.ts
@@ -59,13 +59,13 @@ export const handleKeydown = (key: string, shiftKey: boolean, ctrlKey: boolean, 
 	* When backspace is pressed, if the cursor is inside an empty inline math,
 	* delete both $ symbols, not just the first one.
 	*/
-	if (settings.autoDelete$ && key === "Backspace") {
-		const char_at_pos = getCharacterAtPos(view, ctx.pos);
-		const char_at_prev_pos = getCharacterAtPos(view, ctx.pos - 1);
+	if (settings.autoDelete$ && key === "Backspace" && ctx.mode.inMath()) {
+		const charAtPos = getCharacterAtPos(view, ctx.pos);
+		const charAtPrevPos = getCharacterAtPos(view, ctx.pos - 1);
 
-		if (char_at_pos === "$" && char_at_prev_pos === "$") {
+		if (charAtPos === "$" && charAtPrevPos === "$") {
 			replaceRange(view, ctx.pos - 1, ctx.pos + 1, "");
-			// note: not sure if removeAllTabstops is necessary
+			// Note: not sure if removeAllTabstops is necessary
 			removeAllTabstops(view);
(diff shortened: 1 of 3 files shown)


artisticat1:
Implementation looks fine to me!

I am not sure if this is necessary, but perhaps we should clear all tabstops when this effect is triggered.

bdm-k:
Do you mean calling `removeAllTabstops`, which is defined in "tabstops_state_field.ts"?

artisticat1:
Yeah, exactly.

bdm-k:
Could you tell me specifically what other changes I should make?

artisticat1:
I'm happy to merge after this is done:

> * making this feature toggleable
> * organizing the code into a new file

bdm-k:
I'm having second thoughts. Isn't it a bit excessive to create a new file since there are only a few lines of code?

artisticat1:
Actually, can we check whether `context.mode.inlineMath` is `true` instead of looking for `$` symbols? This would be more semantic as well as robust.

> I'm having second thoughts. Isn't it a bit excessive to create a new file since there are only a few lines of code?

That's true; it would be nice for consistency with the other functions/features though. I'll leave it up to you.

bdm-k:
> Actually, can we check whether `context.mode.inlineMath` is `true` instead of looking for `$` symbols? This would be more semantic as well as robust.

In the case of `$|$`, `context.mode.inlineMath` will be `false`. So, we can't determine whether the cursor is inside an empty inline math or entirely outside of an inline math.

artisticat1:
Oh, that sounds like unintended behaviour - I'll try to look into it.

bdm-k:
When the text is `$$`, the syntax tree will be:
```
Document [1:0..1:2]
   └─ formatting_formatting-math_formatting-math-begin_keyword_math_math-block [1:0..1:2]: "$$"
```

On the other hand, when the text is `$x$`, the syntax tree will be:
```
Document [1:0..1:3]
   ├─ formatting_formatting-math_formatting-math-begin_keyword_math [1:0..1:1]: "$"
   ├─ math_variable-2 [1:1..1:2]: "x"
   └─ formatting_formatting-math_formatting-math-end_keyword_math_math- [1:2..1:3]: "$"
```

In the first case, no node contains "math-end" in its name, so `context.mode.inlineMath` will be `false`. Therefore, we must directly look for `$` symbols.

artisticat1:
Okay, I'll go ahead and merge this then. Thank you!

PR #286, opened 2024-04-24
Add an option to disable snippet expansion when IME is on

### LINKED ISSUE #284: [FEATURE] Disable snippet expansion when IME is on
Thank you for maintaining this important plugin for all who do math in Obsidian. I have a suggestion which will be beneficial for those who use IME.

# Description of the Problem

When I type "Vimキー" (which means "Vim keys" in Japanese), the keystroke looks like this:
```
Shift+V, i, m, (Turn on IME), k, i, -, (space to convert hiragana to katakana)
```
Then, the `m` before turning on the IME and `k` after that triggers the default `mk` snippet, and it results in the following, which is unwanted:
```
Vi$きー$
```

# Description of the Solution

I think this problem can be eliminated by disabling expanding snippets while IME is on, i.e. when [view.composing](https://codemirror.net/docs/ref/#view.EditorView.composing) is `true`.

# Additional

None

RyotaUshio:
Resolves #284 

I have tested it on macOS and iPadOS, Japanese IME only.
If you have any suggestions, please feel free to let me know.

https://github.com/artisticat1/obsidian-latex-suite/assets/72342591/38634f4e-93bd-4f6e-9232-d1cd2561406a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "obsidian-latex-suite",
-	"version": "1.9.1",
+	"version": "1.9.2",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "obsidian-latex-suite",
-			"version": "1.9.1",
+			"version": "1.9.2",
 			"license": "MIT",
 			"dependencies": {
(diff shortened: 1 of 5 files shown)


artisticat1:
Thank you!

PR #288, opened 2024-04-25
Improve tooltip position to avoid hiding the line math section

### LINKED ISSUE #275: [BUG] Inline math popup issue with line wrap
# Description

With math popup preview position set to "Below", when the equation gets long enough to wrap into two or more rows the popup obstructs the view.
# To Reproduce

- set math popup preview position to "Below"
- write inline math long enough that it wraps into multiple rows
# Expected Behavior

If set to "Below", popup should stay under the last row of the inline math section so that all the rows stay visible.

# Screenshots

Popup set to "Above":
(current workaround, but not a solution since I'd like to be able to see previous text)

image

Popup set to "Below":

image

# Platform and versions

Windows 10
Obsidian v1.5.8
Latex Suite 1.9.1

bdm-k:
Closes #275 

image

### DIFF
--- a/src/editor_extensions/math_tooltip.ts
+++ b/src/editor_extensions/math_tooltip.ts
@@ -1,58 +1,110 @@
-import { Tooltip, showTooltip, EditorView } from "@codemirror/view";
-import { StateField, EditorState } from "@codemirror/state";
+import { Tooltip, showTooltip, EditorView, ViewUpdate } from "@codemirror/view";
+import { StateField, EditorState, EditorSelection, StateEffect } from "@codemirror/state";
 import { renderMath, finishRenderMath, editorLivePreviewField } from "obsidian";
 import { Context } from "src/utils/context";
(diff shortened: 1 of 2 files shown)


src/editor_extensions/math_tooltip.ts
```diff
@@ -1,58 +1,121 @@
...
 
 	provide: (f) => showTooltip.computeN([f], (state) => state.field(f)),
 });
 
-function getCursorTooltips(state: EditorState): readonly Tooltip[] {
-	const settings = getLatexSuiteConfig(state);
-	const ctx = Context.fromState(state);
+// update the tooltip by dispatching an updateTooltipEffect
+export function handleMathTooltip(update: ViewUpdate) {
+  const shouldUpdate = update.docChanged || update.selectionSet;
```
artisticat1, line 25: should be indented with 1 tab

artisticat1:
Thank you!

PR #292, opened 2024-05-06
Correctly determine whether the cursor is within a code block

### LINKED ISSUE #290: [BUG] Incorrectly determine whether the cursor is within a code block
If the cursor is placed at the beginning of a line that forms part of a code block, it is considered outside of a code block. The example below illustrates this scenario, with the vertical bar indicating the cursor's position:

> \```typescript
  |console.log("Hello, World");
  \```

https://github.com/artisticat1/obsidian-latex-suite/blob/1d49c1bfd659899db129c64695a224380a335ee4/src/utils/context.ts#L270
Specifically, the `inCodeblock` variable will be `false` in such cases.

bdm-k:
Closes #290


artisticat1:
Thanks!

Issue #303, opened 2024-06-19
Tabout sometimes goes backward to random part of previously written math[BUG] 


alexnegron:
# Description
Tab out sometimes skips to a random math line somewhere else in the document. 

# To Reproduce
Say you're typing text and some math in expression A (either in $$ or $$ $$). Continue writing until you write an expression B. Sometimes, when tabbing while writing expression B, the cursor will jump back to some spot in expression A. 

# Expected Behavior
One would expect the tabbing to proceed forward, one by one, but sometimes it goes backward in an unpredictable way.

bdm-k:
I suspect this bug happens when you move the cursor using a pointing device (mouse or touchpad) instead of a keyboard.

@alexnegron 
Can you find any reproducible examples?

alexnegron:
Thanks for your update!

Issue #254, opened 2024-02-12
[BUG] Un-triggered tabstops create a backlog


git-wil:
# Description
When using snippets with tabstops, it seems like un-triggered tabstops begin to pile up into a queue that can trigger later. For example, if you have a tabstop at the end of an expression but use arrow keys to step past it, pressing tab at any time after the expression will cause the cursor to jump backwards to the tabstop you have already passed.

# To Reproduce
As mentioned above, if you use arrow keys to move the cursor after a tabstop, pressing tab will jump the cursor backwards to the stop. 

# Expected Behavior
I would suggest that, if the cursor's location is past the location of the tabstop, the stop should be removed from the tabstop queue. This would mean there would be no unexpected jumps backwards when pressing tab. If this backwards jumping is considered intended behavior, then at least having a checkbox in the plugin's settings to only allow tab to jump to stops that are in front of the cursor would be a much appreciated capability.

gamberoillecito:
I experience this frequently aswell

superle3:
Fixed by #306

PR #306, opened 2024-06-24
New architecture to manage tabstops

bdm-k:
This PR introduces a new method for managing tabstops. With the updated architecture, we can reduce the chances of bugs described in issues #303 and #254 occurring.

## Changes
Currently, we need to dispatch the dedicated state effects multiple times to update `tabstopsStateField`, which is a bit complicated. However, with the new implementation, we'll only need to dispatch a single transaction that provides the new editor selection. As a result, since the `tabstopsStateField` is updated based solely on editor selection changes, we no longer need to listen to view updates to keep tabstop groups in the correct state.

### Details
* `consumeAndGotoNextTabstop` has been replaced with `setSelectionToNextTabstop`. This new function determines the next tabstop position and dispatches the transaction that changes the editor selection.
* `tidyTabstops` has been removed; its role is now fulfilled by the update logic of the `tabstopsStateField`.
* The process of updating tabstop groups in the `handleUpdate` function has been removed.

### Before

https://github.com/artisticat1/obsidian-latex-suite/assets/54701853/cb6081e1-fb0e-413d-8853-e8a99adf63db

### After

https://github.com/artisticat1/obsidian-latex-suite/assets/54701853/f9a5a7fc-8779-4174-bb74-efa131a493bf


alexnegron:
Excellent. Thank you!

artisticat1:
Thanks for the PR! I'll try to review it in the next week or so.

PR #310, opened 2024-06-30
Allows for matrix commands in inline math

### LINKED ISSUE #262: [FEATURE] Support matrix shortcuts in inline math
# Description of the Problem

Matrix shortcuts don't currently work in inline math, like
```
$\begin{pmatrix}5&6&7\\8&9&10\end{pmatrix}$
```

# Description of the Solution

The same matrix shortcuts should work in inline matrices, with some sort of shortcut to insert `\\` for a new line.

# Additional context

I often write inline matrices when taking notes in class (I find display math takes up too much space) - I modified the default matrix snippets to remove the `\n` so I can use them inline.

AJCubed:
Resolves #262.

Removes the newlines from the matrix snippet expansion while in inline mode, and allows the tab/enter keys insert & or \\  to even if you're inline. Under the new behavior, shift-tab will bring the cursor to the next tabstop, and shift-enter will move the cursor to the next line. 

Demo:
https://github.com/artisticat1/obsidian-latex-suite/assets/64773493/130d8ad9-e9ff-4104-a589-a6af14ee8f2a

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -20,7 +20,7 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, 
 	if (!isInsideAnEnv) return false;
 
 
-	if (key === "Tab") {
+	if (key === "Tab" && !shiftKey) {
 		view.dispatch(view.state.replaceSelection(" & "));
 
 		return true;
@@ -34,6 +34,9 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, 
 			const nextLine = d.line(nextLineNo);
 
 			setCursor(view, nextLine.to);
+		}
+		else if (ctx.mode.inlineMath) {
+			view.dispatch(view.state.replaceSelection(" \\\\ "));
 		}
 		else {
 			view.dispatch(view.state.replaceSelection(" \\\\\n"));
--- a/src/features/run_snippets.ts
+++ b/src/features/run_snippets.ts
@@ -124,6 +124,11 @@ const isOnWordBoundary = (state: EditorState, triggerPos: number, to: number, wo
 }
 
 const trimWhitespace = (replacement: string, ctx: Context) => {
+	// Modify matrix expansion for inline math
+	const regex = new RegExp("\\n\\$0\\n", "g");
+  	replacement = replacement.replace(regex, " $0 "); 
+
+	// Trim whitespace
 	let spaceIndex = 0;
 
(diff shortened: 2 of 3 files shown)


artisticat1:
Thank you for the PR!

src/features/run_snippets.ts
```diff
@@ -124,6 +124,11 @@ const isOnWordBoundary = (state: EditorState, triggerPos: number, to: number, wo
 }
 
 const trimWhitespace = (replacement: string, ctx: Context) => {
+	// Modify matrix expansion for inline math
+	const regex = new RegExp("\\n\\$0\\n", "g");
+  	replacement = replacement.replace(regex, " $0 "); 
+
+	// Trim whitespace
```
artisticat1, line 131: This shouldn't be needed with the current default snippets (https://github.com/artisticat1/obsidian-latex-suite/blob/4021ed5c0305acb60808e5a40f9b4602f1142d0d/src/default_snippets.js#L239-L244).

We can add the ability to remove newlines when expanding snippets inside inline math, but it should probably be separate from the `trimWhitespace` function

src/features/matrix_shortcuts.ts
```diff
@@ -20,7 +20,7 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 	if (!isInsideAnEnv) return false;
 
 
-	if (key === "Tab") {
+	if (key === "Tab" && !shiftKey) {
```
artisticat1, line 23: This currently has no effect ("Tab" is still used to advance to the next tabstop, not "Shift+Tab") because snippet expansion/tabout is handled before matrix shortcuts in `handleKeydown`.

I'd prefer to keep the original behaviour anyway ("Tab" to move to the next tabstop), unless you have an argument against it?

MoisesAdame (approved):

src/features/matrix_shortcuts.ts
AJCubed: If I recall correctly I think changing `settings.matrixShortcutsEnabled && ctx.mode.blockMath` to strictlyInMath made it register the shortcut, but when I revise the code I'll confirm.

Re the second part: I think it makes more sense to have the inline commands identical to the block math commands, where Tab inserts an ampersand, since it'd also be the least cumbersome to type. 

Right now in a block math matrix,  Shift-Enter advances to the next tab stop (unless the closing $$ aren't on their own line), so we could also make Shift-Enter behave the same way inline as it does in block math. Let me know what makes the most sense here.

Issue #320, opened 2024-08-08
[BUG] LaTeX Suite will significantly affect Obsidian’s performance.


CamWam:
# Description
I once posted this https://forum.obsidian.md/t/i-have-tested-the-13900k-3090-7950x-4090-and-7950x3d-4090-and-there-is-a-serious-issue-with-canvas-performance-no-configuration-can-smoothly-run-a-canvas-card-with-26-000-characters-of-content/81855/19
on the Obsidian forum. After testing, I found that the input lag in canvas and markdown caused by plugins is mainly due to LaTeX Suite.

Generally, when the input in canvas cards and markdown exceeds 10,000～25,000 characters (markdown documents can be higher), enabling the LaTeX Suite plugin will cause lag in English, Chinese Pinyin, and LaTeX input. Note that this input lag also occurs with regular Chinese and English input. Because the $ ... $ mode cannot use the tab key to automatically input &, I mainly use version 1.7.4

[inputting.zip](https://github.com/user-attachments/files/16538162/inputting.zip)

HerrChaos:
This is still happening and really bad. Typing in any kind of LaTeX block start to lag and I need to wait at least half a second for the character to appear or for the cursor to move. Typing outside the LaTeX block works fine but inside it lags bad. 
I disabled Latex and the lag stopped. The note I typed in had only around 1,4k characters and most of them in LaTeX blocks. Although I have experience this lag in smaller documents too. My current LaTeX suit version is 1.9.8. 
I love this plugin and would love to get a performance updates since it makes math almost unbearable to type.

TLDR. Lags badly in LeTeX blocks. Pls fix.

HerrChaos:
Ok, so I fixed the lags I noticed myself and also I have to say it was not as bad as I thought. The main issue came from the Copilot plugin. Still not good but not as bad as I thought.

superle3:
Has this been improved as of 1.11? as I think the most improvements that can be made have been made (without major rewriting/ introducing a new parser)

superle3:
Closing due to inactivity and think most problems are solved

PR #323, opened 2024-08-25
Improve performance of `runSnippetCursor`

shunby:
Optimizes `runSnippetCursor` in `run_snippet.ts` by reducing redundant calls to `view.state.sliceDoc`.
Partially solves #320 by improving the editting of large files.

# Performance Testing

Tested with a large markdown file generated with the following command:
```bash
python -c "print('aa\n'*4000)" > test.md
```
Pushed a key when the cursor is at the bottom of the file and measured the time spent handling the keydown event using DevTools.

Test Environment:
- OS: Microsoft Windows 11 Home
- CPU: Intel(R) Core(TM) i7-8750H CPU
- RAM: 16GB

Results:
- Before the update: Handling a single keydown event took approximately 50ms, with `runSnippetCursor` consuming 47ms.
- After the update: Handling a single keydown event took approximately 3ms, with `runSnippetCursor` consuming 0.2ms.

### DIFF
--- a/src/features/run_snippets.ts
+++ b/src/features/run_snippets.ts
@@ -34,9 +34,10 @@ const runSnippetCursor = (view: EditorView, ctx: Context, key: string, range: Se
 	const settings = getLatexSuiteConfig(view);
 	const {from, to} = range;
 	const sel = view.state.sliceDoc(from, to);
-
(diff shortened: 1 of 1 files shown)


artisticat1:
Thanks!

PR #332, opened 2024-09-27
feat: Delaying the revealment of concealments

### LINKED ISSUE #283: [FEATURE] Delaying the reveal of concealed math
# Description of the Problem

When moving the text cursor left or right through an equation with conceal mode turned on, it's impossible to know how many clicks you need to select exactly what you want or to position the cursor where you want. If I see I'm three characters from the place I want to get to, if it's concealed math then I'm further away. Real LaTeX syntax will be revealed every time I get to a new symbol and be hidden once I pass it, which is very annoying.

# Description of the Solution

A time delay, where in the first `x` seconds a symbol of concealed math that I'm next to won't be replaced with real underlying LaTeX.

1s seems good to start with, but a configurable option would be the best.

That way it is possible to edit equations faster, more predictable, and easier - not requiring the user to think about syntax when editing and letting him focus on math.

# Additional information

When a section of text is selected, the symbols should convert to LaTeX syntax after the time delay, and should keep the accurate selection over themselves, if possible.

If the text cursor is positioned between two "concealed math" symbols and the user starts typing, the symbols won't convert to real syntax until the cursor has been idle for longer than `x`.

# Open for suggestions

If anyone who reads this finds that it would make more sense to add/modify a part of the idea, please comment so we can discuss it and find the best solution. :)

bdm-k:
Closes #283 

### Key points of the implementation
We introduced 3 new types: `Replacement`, `ConcealSpec`, and `Concealment`. It's a bit confusing, but `Replacement` is equivalent to `Concealment` in the previous code, while the new `Concealment` now refers to something different.

- `ConcealSpec`: array of `Replacement`. Replacements in the same conceal spec are enabled or disables simultaneously. This layer is necessary due to LaTeX commands such as `\frac` and `\set`.
- `Concealment`: `ConcealSpec` with additional context. The decorations and atomic ranges are built from a list of `Concealment` objects.

The `conceal` function collects conceal specs. Note that it does not rely on the mouse position or whether the mouse is down. The conceal specs collected are converted into `Concealment` objects using information such as the previous `Concealment` objects. At this time, the ones that should be revealed after a delay are passed to the [debouncer](https://docs.obsidian.md/Reference/TypeScript+API/debounce).


artisticat1:
Thanks for the PR! I'll take a look soon.

artisticat1:
This works very well. Thanks!

Brenzle:
This is awesome, thanks! All of you!

PR #343, opened 2024-11-07
Add gather to default env names for matrix shortcuts

llakala:
I was confused why the matrix shortcuts commands weren't working in the gather environment. It turns out, gather just wasn't in the default list for environments to enable to shortcuts.


artisticat1:
Thanks!

PR #344, opened 2024-11-07
Conceal all types of fractions, not just \frac

llakala:
Currently, the only fractions that are concealed are `\frac`. I prefer using `\dfrac` for everything, since it'll always display as large, even when doing inline math. However, doing this has made me lose out on concealed fractions.This PR aims to fix that. Now, the regex for finding fractions looks for all types of fractions. 

Regex is absolutely not my strong suit, and I wouldn't be surprised if there's a better way to match for the string "frac" that *may* or may not be preceded by `d`, `t`, or `g`. Feel free to enlighten me on this!

### DIFF
--- a/src/editor_extensions/conceal_fns.ts
+++ b/src/editor_extensions/conceal_fns.ts
@@ -375,7 +375,7 @@ function concealSet(eqn: string): ConcealSpec[] {
 function concealFraction(eqn: string): ConcealSpec[] {
 	const concealSpecs: ConcealSpec[] = [];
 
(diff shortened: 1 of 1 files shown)


artisticat1:
Thanks!

> Regex is absolutely not my strong suit, and I wouldn't be surprised if there's a better way to match for the string "frac" that *may* or may not be preceded by `d`, `t`, or `g`. Feel free to enlighten me on this!

`(d|t|g)?frac` will work as you described :)

Your current solution also works well, so I'm happy to merge!

PR #347, opened 2024-11-10
Manage symbols exceptions for auto-subscript

### LINKED ISSUE #346: [FEATURE] Insert space after some symbols
# Description of the Problem
Discussed in https://github.com/artisticat1/obsidian-latex-suite/discussions/337
Greater/less than/equal, to, ln, log, trig... symbols automatically default to a subscript for the following number

# Description of the Solution
Update src/default_snippets.js
`{trigger: "\\\\(${MORE_SYMBOLS}|${SYMBOL})(.)", replacement: "\\[[0]] [[1]]", options: "rmA"}`
add `mid`, `lt`, `gt`, `le`, `ge` and trigs fonctions to src/default_snippet_variables.js

Captainbleu:
close #346

### DIFF
--- a/src/default_snippets.js
+++ b/src/default_snippets.js
@@ -50,6 +50,7 @@
 	{trigger: "ee", replacement: "e^{ $0 }$1", options: "mA"},
     {trigger: "invs", replacement: "^{-1}", options: "mA"},
     {trigger: /([A-Za-z])(\d)/, replacement: "[[0]]_{[[1]]}", options: "rmA", description: "Auto letter subscript", priority: -1},
+	{trigger: "\\\\(${MORE_SYMBOLS}|${SYMBOL})(.)", replacement: "\\[[0]] [[1]]", options: "rmA", description: "Manage symbols exceptions"}
 
     {trigger: /([^\\])(exp|log|ln)/, replacement: "[[0]]\\[[1]]", options: "rmA"},
     {trigger: "conj", replacement: "^{*}", options: "mA"},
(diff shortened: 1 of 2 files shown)


artisticat1:
Thank you for the PR!

src/default_snippets.js
```diff
@@ -50,6 +50,7 @@
 	{trigger: "ee", replacement: "e^{ $0 }$1", options: "mA"},
     {trigger: "invs", replacement: "^{-1}", options: "mA"},
     {trigger: /([A-Za-z])(\d)/, replacement: "[[0]]_{[[1]]}", options: "rmA", description: "Auto letter subscript", priority: -1},
+	{trigger: "\\\\(${MORE_SYMBOLS}|${SYMBOL})(.)", replacement: "\\[[0]] [[1]]", options: "rmA", description: "Manage symbols exceptions"},
```
artisticat1, line 53: Do we need to capture all characters here? Can we apply the fix only to numbers typed after a symbol?

Also, do we want make "log" an exception to the rule (i.e. continue to apply auto-subscript to it), so we can type e.g. `log2` -> `\log_{2}`?

src/default_snippets.js
```diff
@@ -50,7 +50,7 @@
 	{trigger: "ee", replacement: "e^{ $0 }$1", options: "mA"},
     {trigger: "invs", replacement: "^{-1}", options: "mA"},
     {trigger: /([A-Za-z])(\d)/, replacement: "[[0]]_{[[1]]}", options: "rmA", description: "Auto letter subscript", priority: -1},
-	{trigger: "\\\\(${MORE_SYMBOLS}|${SYMBOL})(.)", replacement: "\\[[0]] [[1]]", options: "rmA", description: "Manage symbols exceptions"},
+	{trigger: "\\\\((${MORE_SYMBOLS}|${SYMBOL})^(log))(.)", replacement: "\\[[0]] [[1]]", options: "rmA", description: "Manage symbols exceptions"},
```
Captainbleu, line 53: Like you want, here it's the version with the log exception

src/default_snippets.js
Captainbleu: Like you want. I done a new commit for this exception.

Issue #44, opened 2022-07-26
VISUAL with vim shortcuts


mscott99:
I like the VISUAL idea for editing previously written latex. However it is easier to select text while in vim mode. Would there be a way to use the selection performed in vim mode?

I am not sure of what exactly this would look like.

Best,

Matthew

artisticat1:
Yep, this is definitely possible. I'll try to fix this in one of the next releases.

rwang2022:
I think it's a great idea. Are there any plans to implement this soon? 

The issue, I think comes from the fact that vim doesn't really support doing a visual selection then going into insert mode (without losing the selection), so right now you can only select normally using mouse or shift+arrow keys. Not exactly sure what a workaround would look like.

Best, 
Ryan

superle3:
its possible using the vimrc plugin (https://github.com/esm7/obsidian-vimrc-support/tree/master) in a kinda hacky way.
by adding 
```vim
exmap test jscommand {const sel = selection; editor.cm.cm.state.vim.insertMode = true; editor.setSelection(sel.anchor, sel.head);}
vmap <C-o> :test<CR>
```
to your vimrc you should be able to use the visual snippets.
This does hide the cursor, but I couldn't figure out from the docs where the actual vim functions are in the editor object.
it would probably be easier to add 2 commands/shortcuts to switch between insert select mode and visual select mode in the obsidian plugin itself

superle3:
adding

```js
this.addcommand({
	// this adds the select mode from vim so visual snippets can be triggered with vim visual selection
	id: "switch-to-select-mode",
	name: "Switch from visual mode to select mode",
	editorCheckCallback: (checking: boolean, editor: Editor) => {
		let vimObject = (window as any).CodeMirrorAdapter?.Vim;
		if (
			!(this.app as any).isVimEnabled() ||
			!editor.cm.cm.state.vim?.visualMode
		) {
			return false;
		} else if (!checking) {
			let selection: EditorSelection = editor.listSelections();
			vimObject.handleKey(editor.cm.cm, "<Esc>");
			vimObject.handleKey(editor.cm.cm, "i");
			editor.setSelections(selection);
		} else {
			return true;
		}
	},
});
```
would add select mode as an command. you can go back to visual with <C-o> (vim mapping).
This does work with multicursor. this doesn't give an indirect mapping sadly, cause : in visual -> :'<,'>
or for only vimrc
```js
const sel = selection;
[...]

Issue #352, opened 2024-11-17
[FEATURE] Intercept Vim-triggered newlines for the Matrix environment `\\` shortcut


llakala:
# Description of the Problem
Currently, within a Matrix environment, a `\\` symbol is inserted whenever you press Enter. However, if you have have Obsidian's Vim mode enabled, and press `o` to add a newline and enter Insert mode, no `\\` is added. 

# Description of the Solution
Hypothetically, the plugin would be able to detect the Vim action and add `\\` when pressing `o`, just like when pressing Enter normally. 

I don't know if this is *possible*. If the newline isn't being detected by `KeyboardEvent`, I assume that Obsidian's does something like this when pressing `o`:
1) add the newline through its own systems, undetectable to us
2) enter Insert mode, where we can now detect inputs (but a bit late for that)

If it *is* possible to hook into this mechanism, it's not obvious to me how one would do it. I originally planned on making a PR to implement this, but realized I didn't know what direction to go in. I'd love some feedback on whether people think this would be possible, and if so, how it could be implemented.

llakala:
No luck after some hacking away at this. I was hoping that this line would work: 
```vim
map o g$a<CR>
```

However, it seems that Obsidian doesn't recognize <CR> as pressing Enter, which is what [Matrix shortcuts expect to insert `\\`](https://github.com/artisticat1/obsidian-latex-suite/blob/ba17d5cd47b4e94a7aba7add32e98f575480a322/src/features/matrix_shortcuts.ts#L28). This would work perfectly if either: 
- The matrix shortcuts would be able to accept `<CR>` as a valid newline somehow. Currently, the plugin just taps into keybinds, so I assume this would require extra functionality to tap into what Vim sends to Obsidian. 
- Some form of vim shortcut could somehow manually trigger the literal Enter symbol rather than `<CR>`, which the matrix shortcuts would then accept.

This is where I hit a wall when it came to knowledge, and searching didn't help much. If anyone has any knowledge in either of these areas, please let me know.

superle3:
```js
export function getVimSpecialEnterCommand(settings: LatexSuitePluginSettings): vimCommand {
	return {
		id: "latex-suite-vim-special-enter",
		defineType: "defineAction",
		type: "action",
		action: (cm: CodeMirrorEditor) => {
			//@ts-ignore
			const vimObj: Vim | null = window?.CodeMirrorAdapter?.Vim;
			if (!vimObj) return;
[...]

PR #359, opened 2024-12-05
Enable visual snippets for vim

superle3:
This PR tries to somewhat implement https://github.com/artisticat1/obsidian-latex-suite/issues/44 by adding a keybinding to swap between `keymap` `vim-insert` and `vim-visual` without losing the current selection. 
It's implemented as a setting instead of an hotkey such that more complex commands like <C-g><C-g> are possible.
The setting is turned off by default and can only be turned on if obsidians vim mode is on.

Its referred to as SelectMode in the settings, because vim's selectmode is something similair to selecting in `vim-insert` in codemirror, but not the same. For example  visualblock and visualline are not retained after switching only the selection and maybe more things, but just tries to expose the `vim-insert` keymap while in `vim-visual`, not emulating vim's actual selectmode.

The vim object is behind untyped objects from obsidian so I copied the types from the original codemirror vim project (https://github.com/replit/codemirror-vim), but these have a lot of any in them so its not that great.

Also this may not be the right plugin for this PR, I understand if its better suited for https://github.com/esm7/obsidian-vimrc-support/tree/master or the vim extension itself.

at last, first time using ts/js, sorry for potential bad code quality


llakala:
Tested this, and can confirm basic functionality is working. I can't speak for code quality, but I'd love to see something added to `DOCS.md` about this before merging.

Also, for what it's worth, I personally think this fits in perfectly fine into this plugin, and doesn't need to be put into something else.

superle3:
looked a bit more into unmap and map and changed descriptions accordingly. Also added ? for the undocumented objects, since im just copying that from the vimrc plugin and I have no idea what their behavior is. Also removed setting keyword from settings tab to comply with obsidian guidelines. But behavior is the same  so amended it.

Okay after checking `<S-a>` mapping again, it seems shift doesn't work in vim mappings as in the actual shift key `<S-[key]>` not `<C-a>B`.  Will report it to the vim repo, but will change the description  accordingly again since the change will take time.

superle3:
also would like to add https://github.com/artisticat1/obsidian-latex-suite/issues/352. Would do it in a seperate PR, but I wouldn't be able to reuse my current code for that

llakala (changes requested):
Just tested, enter shortcut is working beautifully. Thanks for your work!

DOCS.md
```diff
@@ -280,3 +281,26 @@ You can [view snippets written by others and share your own snippets here](https
...
+
+### Select mode
+You can trigger visual snippets in insert mode, but not in visual mode.
+Because of this, a shortcut in the vim extension is defined that allows you to switch between insert mode and visual mode while keeping the selection. This should also work with multicursor.
+
+You can define this shortcut in the settings by enabling vim mode in obsidian and in this plugin.
+The shortcut should be a vim keybinding with no spaces. Some keybindings are not supported by the original vim extension. Due to this please first try seeing if the shortcut works for other keys like `:imap <C-a> <Esc>` or `:vmap <A-h> 0`. It is known that the shift key like `:map <S-a> <Esc>` does not work.
```
llakala, line 299:
```suggestion
The shortcut should be a vim keybinding with no spaces. Some keybindings are not supported by the original vim extension. Due to this, please first see if the shortcut works for other keys, such as `:imap <C-a> <Esc>` or `:vmap <A-h> 0`. It is known that the shift key (`:map <S-a> <Esc>`) does not work.
```

DOCS.md
```diff
@@ -280,3 +281,26 @@ You can [view snippets written by others and share your own snippets here](https
...
+
+The shortcut from visual mode to select mode only works in visual mode and the shortcut from select mode to visual mode only works when something is selected and you are in insert mode.
+
+Macros are not yet supported for visual snippets
```
llakala, line 303:
```suggestion
Macros are not yet supported for visual snippets.
```

llakala (changes requested):
Small wording changes to new documentation. Newline logic working perfectly - I'll spend some time using this PR on my main vault to see if I encounter any bugs.

superle3:
Now `o` doesn't work outside of environments. Which makes sense from the code, but I thought I tested that and it did work.
easy solution would be to add 
```ts
if (!runMatrixShortcuts(view, ctx, "Enter", false)) {
    cm.getInputField().dispatchEvent(new KeyboardEvent("keydown", { key: "Enter" }));
}

```
but that changes the behaviour of `o` for empty indented lines.
I tried to look at the original code comes down to the code below, but I am not really sure whats going on there. I would have to look at the commands api. 
Most importantly though the current implementation from both vim and codemirror-vim seem counterintuitive (although vim probably makes some sense).

I would expect it to keep the current indentation regardless if the line is empty, but
codemirror-vim: uses the current indentation for the the newline but deletes the current indentation if the line is empty.
vim (tested for python on `vim --clean`): finds the proper indentation to use for the newline through the parser (did not work for md file) and deletes the current indentation if the line is empty and has not been edited so if you do the sequence `o<Esc>o`.

The code below does work instead of pressing the enter key.
```ts
import {insertNewlineAndIndent} from "@codemirror/commands"
    newlineAndIndent: function (cm: CodeMirror) {
      insertNewlineAndIndent({
        state: cm.cm6.state,
        dispatch: (tr) => {
          return dispatchChange(cm, tr);
        }
      })
function dispatchChange(cm: CodeMirror, transaction: any) {
  var view = cm.cm6;
  if (view.state.readOnly)
    return;
  var type = "input.type.compose";
  if (cm.curOp) {
    if (!cm.curOp.lastChange) type = "input.type.compose.start";
  }
  if (transaction.annotations) {
    try {
      transaction.annotations.some(function (note: any) {
        if (note.value == "input") note.value = type;
      });
    } catch (e) {
      console.error(e);
    }
  } else {
    transaction.userEvent =  type;
  }
  return view.dispatch(transaction)
}
```

superle3:
I couldn't find out the reason for this code from the documentation of both codemirror and the vim plugin. There is a slight change from the code a both. insertmode is entered after a newline has been added since otherwise macros would behave weird and the annotations are only changed if its an array. I don't know why there is a try catch but im guessing it has a reason.

superle3:
forgot to unmap the old keymappings when vim keybindings are turned off. Everything should be good now

llakala (approved):
Newline behavior working perfectly for me, thank you!

superle3:
For the type safety it would probably be better to use https://github.com/Fevol/obsidian-typings for the untyped objects. But the methods of the `vimApi` is still untyped so its still not ideal.

superle3:
changed `window.CodeMirrorAdapter.Vim` with `initVimMode(CodeMirrorAdapter)`

PR #362, opened 2024-12-28
Inline Matrix Commands

AJCubed:
Replaces PR #310 , which was accidentally deleted. 

As the previous discussion outlined, we now have the following behavior in inline math:
Tab: inserts an ampersand (&)
Enter: inserts " \\\\ "
Shift-Enter: tabs out of the matrix environment.

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -2,7 +2,7 @@ import { EditorView } from "@codemirror/view";
 import { setCursor } from "src/utils/editor_utils";
 import { getLatexSuiteConfig } from "src/snippets/codemirror/config";
 import { Context } from "src/utils/context";
-
+import { tabout } from "src/features/tabout";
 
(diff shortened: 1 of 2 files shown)


AJCubed:
I did notice a quirk that within a matrix environment (in both the original block behavior and new inline behavior), the `\frac{a}{b}` lets you tab through the segments when first typing it, but when you click back in and then hit Tab it'll insert an ampersand instead of jumping tabstops. This doesn't happen with other bracketed environments like `\text{}` or `a_{b}` or nested matrices.

artisticat1:
Thanks!

> I did notice a quirk that within a matrix environment (in both the original block behavior and new inline behavior), the `\frac{a}{b}` lets you tab through the segments when first typing it, but when you click back in and then hit Tab it'll insert an ampersand instead of jumping tabstops. This doesn't happen with other bracketed environments like `\text{}` or `a_{b}` or nested matrices.

Hmm, that's not intentional. I think

Issue #363, opened 2024-12-30
[BUG] indent and unindent with tab completely impossible


dtinhvo:
# Description
When this plugin is enabled, tab to indent and shift-tab to indent in math block is completely impossible with no known alternative. 

# To Reproduce
Enable the plugin, type some math block. Now press tab with the math block content selected, individual lines selected, or at beginning of lines

# Expected Behavior
at least with a selection, tab should indent the line, or block - very few cases are exceptions to this behavior. I understand that tab stops and such might interfere, but since VISUAL is a feature in this plugin, tab and shift-tab should be given priority to this behavior. Or at least a toggle in settings. Current implementation of VISUAL also does not help me with an alternative, because line-wise addition is not supported (which will at least make a snippet for tab that indents).

dtinhvo:
Some other suggestions that is possible to achieve indentable math
- add a command in Command Palette that disables tabout. After this I can select the equation block, press tab to indent each line, then turns on tabout again. The command can be something like this (I havent looked at the source code)
```
this.addCommand({
  id: 'toggle-tabout',
  name: 'Toggle Tabout',
  callback: async () => {
    this.settings.tabout = !this.settings.tabout;
    console.log(`Tabout is now ${this.settings.tabout ? 'enabled' : 'disabled'}`);
    await this.saveSettings();
  },
});
```
- Command to indent or unindent a block code selection, similar to the "Box current equation"

dtinhvo:
For now, a workaround (that I use) is to exploit the visual function snippet function within obsidian-latex-suite itself:

```
    {trigger: "<", replacement: (sel) => {
        let lines = sel.split('\n');
        return lines.slice(0, -1).map(line => line.startsWith('    ') ? line.substring(4) : line) .concat(lines[lines.length - 1]) .join('\n');
    }, options: "mAv"},
    {trigger: ">", replacement: (sel) => {
        let lines = sel.split('\n');
        return lines.slice(0, -1) .map(line => '    ' + line) .concat(lines[lines.length - 1]) .join('\n');
    }, options: "mAv"},
```
This will match Obsidian's  selection pattern when you click on a math block, and indent the content lines between 2 $$ with `>` and vice-versa with `<`
![image](https://github.com/user-attachments/assets/e8138704-8cea-4358-b736-ef4c3efb6439)
![image](https://github.com/user-attachments/assets/d228952b-139a-4779-b955-ade4985d02b4)

superle3:
[...]

PR #365, opened 2025-01-04
Restore tabindent for selections

superle3:
Implement https://github.com/artisticat1/obsidian-latex-suite/issues/363.

Skip the matrixshortcut and the tabout if something is selected since those actions don't make much sense for selections and indenting would be more reasonable.

Weird behaviour in multicursor is expected but the same weird behaviour is also there without this change.
`ctx.mode` gets the main/last cursor so the main cursor is checked for selections.
Also the weird behaviour is hard to trigger in multicursor.

I don't think its relevant but the weird behaviour is when for example you press tab in the following text
```tex
$$
a|
$$

some text |
```
where | is the cursor.

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -19,8 +19,8 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, 
 
 	if (!isInsideAnEnv) return false;
 
-
-	if (key === "Tab") {
+	// Take main cursor since ctx.mode takes the main cursor, weird behaviour is expected with multicursor because of this.
+	if (key === "Tab" && view.state.selection.main.empty) {
 		view.dispatch(view.state.replaceSelection(" & "));
 
 		return true;
--- a/src/latex_suite.ts
+++ b/src/latex_suite.ts
@@ -98,7 +98,10 @@ export const handleKeydown = (key: string, shiftKey: boolean, ctrlKey: boolean, 
 	}
 
 	if (settings.taboutEnabled) {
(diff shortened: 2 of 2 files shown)


artisticat1:
Thank you for the PRs! I'll look into reviewing them soon.

src/features/matrix_shortcuts.ts
```diff
@@ -19,8 +19,8 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 
 	if (!isInsideAnEnv) return false;
 
-
-	if (key === "Tab") {
+	// Take main cursor since ctx.mode takes the main cursor, weird behaviour is expected with multicursor because of this.
+	if (key === "Tab" && view.state.selection.ranges[view.state.selection.mainIndex].empty) {
```
artisticat1, line 23: Use `view.state.selection.main.empty` instead.

src/latex_suite.ts
```diff
@@ -98,7 +98,10 @@ export const handleKeydown = (key: string, shiftKey: boolean, ctrlKey: boolean,
 	}
 
 	if (settings.taboutEnabled) {
-		if (key === "Tab" || shouldTaboutByCloseBracket(view, key)) {
+		// check if the main cursor has something selected since ctx.mode only checks the main cursor.
+		//  This does give weird behaviour with multicursor.
+		if ((key === "Tab" && view.state.selection.ranges[view.state.selection.mainIndex].empty)
```
artisticat1, line 103: Same as above.

superle3:
Done

artisticat1:
Thanks!

PR #374, opened 2025-01-16
Enhance Tabout: Expand delimiters, refactor constants, and add LaTeX \right support.

Testudinidae:
- Expand Tabout support for more brackets.
- Refactor delimiter variables into constants for better maintainability.
- Add handling for LaTeX `\right` command.


artisticat1:
Thank you, this looks good to me overall!

src/features/tabout.ts
```diff
@@ -3,8 +3,82 @@ import { replaceRange, setCursor, getCharacterAtPos } from "src/utils/editor_uti
...
+	"(",
+	"[", "\\lbrack",
+	"{",
+	"\\{", "\\lbrace",
+	"\\langle",
+	"\\lceil", "\\lfloor",
+	"\\vert",
+	"\\|", "\\Vert",
+	"$"
+].sort((a, b) => b.length - a.length);
```
artisticat1, line 18: I don't see this used anywhere; if it isn't needed, I'd prefer to remove it.

Testudinidae: I was originally considering whether a reverse tabout feature could be implemented. Since it hasn’t been developed yet, it does make sense to remove it. Thank you!

Testudinidae:
> Thank you, this looks good to me overall!
> 
> Can you explain what `matchRightCommand()` does and/or what handling the `\right` command means?

Sure.
Sometimes users use special symbols for closing brackets that make it hard to determine the correct closing point. For example, the same symbol `\vert` can be used as both an opening and closing bracket for absolute values, which makes it impossible to handle perfectly. One option could be allowing users to configure which symbols should be treated as closing brackets, but that's a separate discussion.

What we can handle here is that `\right` is always followed by a closing bracket. To illustrate a more extreme case, some regions use unconventional notation for intervals, like `[a, b)` being written as `[a, b[`. (See [Interval Notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints) for reference.)

Testudinidae:
I just rebased to version 1.9.8.

src/features/tabout.ts
artisticat1: Indeed, a reverse tabout feature (probably bound to <kbd>Shift</kbd>+<kbd>Tab</kbd>) would make sense. We can add this back if/when it's developed!

artisticat1:
> Sometimes users use special symbols for closing brackets that make it hard to determine the correct closing point. For example, the same symbol `|` can be used as both an opening and closing bracket for absolute values, or as a middle separator in sets, which makes it impossible to handle perfectly.
> ...

Thanks, makes sense now!

src/features/tabout.ts
```diff
@@ -3,8 +3,75 @@ import { replaceRange, setCursor, getCharacterAtPos } from "src/utils/editor_uti
...
+	"\\{", "\\}", "\\lbrace", "\\rbrace",
+	"<", ">", "\\langle", "\\rangle", "\\lt", "\\gt",
+	"\\lfloor", "\\rfloor", "\\lceil", "\\rceil",
+	"/", "\\\\", "\\backslash",
+	"|", "\\vert",
+	"\\|", "\\Vert",
+	"\\uparrow", "\\downarrow", "\\Uparrow", "\\Downarrow",
+	"\\ulcorner", "\\urcorner",
+	"."
+];
```
artisticat1, line 31: Can we add `>` and `|` to `CLOSING_BRACKETS` to maintain the original behaviour? Additionally, can we align the order of all the delimiters in `CLOSING_BRACKETS` and `DELIMITERS`?

```suggestion
const CLOSING_BRACKETS = [
	")",
	"]", "\\rbrack",
	"}",
	"\\}", "\\rbrace",
	">", "\\rangle",
	"|", "\\vert",
	"\\|", "\\Vert",
	"\\rfloor",
	"\\rceil",
	"\\urcorner",
	"$"
];
const DELIMITERS = [
	"(", ")",
	"[", "]", "\\lbrack", "\\rbrack",
	"\\{", "\\}", "\\lbrace", "\\rbrace",
	"<", ">", "\\langle", "\\rangle", "\\lt", "\\gt",
	"|", "\\vert",
	"\\|", "\\Vert",
	"\\lfloor", "\\rfloor",
	"\\lceil", "\\rceil",
	"\\ulcorner", "\\urcorner",
	"/", "\\\\", "\\backslash",
	"\\uparrow", "\\downarrow", "\\Uparrow", "\\Downarrow",
	"."
];
```

Testudinidae:
I think I should explain the fix: "Fix bug: prevent partial command matches during parsing" (f84cd22).

The simplest explanation is that it prevents the program from interpreting `\rightarrow` as `\right` + arrow, which could cause the cursor to jump in between `\right`|arrow.

I considered that if a user customizes LaTeX commands using the Extended MathJax plugin, they might face more similar conflicts. In the future, I may develop a feature that allows users to define their own delimiters. That's why I added checks like isCommandEnd.

src/features/tabout.ts
Testudinidae: You can also refer to the branch I created:
https://github.com/Testudinidae/obsidian-latex-suite/tree/feature/user-input-closing-brackets
https://github.com/Testudinidae/obsidian-latex-suite/pull/1/files

artisticat1: Allowing the user to define their own delimiters sounds fine to me. Feel free to merge that commit into this branch!

> The > and | symbols have different meanings in various contexts—they can be used as opening or closing brackets, but they also represent "greater than" and "separator" in other cases. In fact, I even removed \vert and \Vert and replaced them with \rvert and \rVert.

I'd rather be able to tabout to > or | in the event they're used as closing brackets than not be able to tabout to them at all. Making the setting configurable does solve this issue though.

Testudinidae:
There are three features of tabout that I would like to suggest for improvement:

1. After tabout, if there is a space after the closing bracket, it would be more convenient if the cursor moved past the space. This would enhance the editing experience, as users who leave spaces likely intend to continue editing beyond them. For commands like `\rangle`, a space is typically required before a non-symbolic character (e.g., an alphanumeric identifier or operator), so if a space is already present, skipping past it would make editing more efficient. To accommodate different user preferences, it might be beneficial to offer this behavior as an optional setting.
2. If there is no closing bracket, the expected behavior could be: first, jumping to the end of the equation, and if tabout is used again, moving outside the equation.
3. Currently, tabout automatically adds a new line when moving outside an equation. While I personally find this feature very useful, I understand that it might lead to unintended behavior for some users. It may be helpful to provide this as an optional setting, allowing users to enable or disable it based on their workflow.

superle3:
Hi @Testudinidae ,

I'm going to start merging my prs in 2 weeks if I don't hear from you back which might cause merge conflicts. I understand that pr has been open for a long time which is why I am waiting about a month before I start merging my prs.

src/features/tabout.ts
```diff
@@ -1,33 +1,162 @@
...
+			}
+
+			continue;
 		}
-		else if (text.slice(i, i + rangle.length) === rangle) {
-			setCursor(view, i + rangle.length);
 
-			return true;
+		// Attempt to match only the right command if matching right command + delimiter fails
+		const rightCommandLength = findTokenLength(SORTED_RIGHT_COMMANDS, text, i);
```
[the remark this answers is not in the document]
superle3: findtokenLength? I'll check tmr

src/settings/settings_tab.ts
```diff
@@ -340,6 +340,19 @@ export class LatexSuiteSettingTab extends PluginSettingTab {
...
+			.setName("Closing brackets")
+			.setDesc("A list of closing brackets for tabout, separated by commas.")
+			.addText(text => text
+				.setPlaceholder(DEFAULT_SETTINGS.taboutClosingSymbols)
+				.setValue(this.plugin.settings.taboutClosingSymbols)
+				.onChange(async (value) => {
+					this.plugin.settings.taboutClosingSymbols = value;
+
 					await this.plugin.saveSettings();
 				}));
```
superle3:
```suggestion
		const taboutClosingBracketsSetting = new Setting(containerEl)
			.setName("Closing brackets")
			.setDesc("A list of closing brackets for tabout, separated by commas.")
			.addText(text => text
				.setPlaceholder(DEFAULT_SETTINGS.taboutClosingSymbols)
				.setValue(this.plugin.settings.taboutClosingSymbols)
				.onChange(async (value) => {
					this.plugin.settings.taboutClosingSymbols = value;

					await this.plugin.saveSettings();
				}));
		taboutClosingBracketsSetting.settingEl.toggleClass("hidden", !this.plugin.settings.taboutEnabled);
		taboutEnabledSetting.addToggle((toggle) => toggle
			.setValue(this.plugin.settings.taboutEnabled)
			.onChange(async (value) => {
				this.plugin.settings.taboutEnabled = value;
				taboutClosingBracketsSetting.settingEl.toggleClass("hidden", !value);
				await this.plugin.saveSettings();
			})
		);
```
something like this (vscode doesn't let me select the whole section for some reason and don't want to figure out how for now)

src/features/tabout.ts
```diff
@@ -1,33 +1,161 @@
...
+const findTokenLength = (text: string, startIndex: number, sortedTokens: string[]): number => {
+	const matchedToken = sortedTokens.find((token) => isMatchingToken(text, token, startIndex));
+
+	if (matchedToken) {
+		return matchedToken.length;
+	}
 
+	return 0;
+}
+
```
superle3: Sorry for the late comment, the "complexity" of the code above bothered me a bit before, not that its complex, but it seemed to be doing more than necessary and I couldn't point out why.

Is there a reason you're not using positive lookahead/regex? I'd think that be more cleanly (unlike positive lookbehind, lookahead is supported on all platforms that obsidian supports)

Its fine if you think this is better, just interesting in hearing your thoughts.

superle3:
@Testudinidae I changed `doc.slice(..).toString` to `doc.sliceString`, merged the 3 transactions at the end into 1 for style and performance reasons and remove the edge case for `\` as `$\$` is not a full latex string because `$` then gets escaped. Thx for simplifying the tokenizer (which I probably to do for the other stuff as we handparse a lot things in weird ways).
Can I merge it now?

PR #375, opened 2025-01-16
Enhance Matrix Shortcuts: Auto-indent for Newlines

Testudinidae:
Enhances the runMatrixShortcuts functionality by adding auto-indentation for newlines. The new feature ensures that any new line created in the editor aligns with the indentation level of the current line, improving usability and maintaining consistent formatting.

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -4,14 +4,14 @@ import { getLatexSuiteConfig } from "src/snippets/codemirror/config";
 import { Context } from "src/utils/context";
 import { tabout } from "src/features/tabout";
 
-export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, shiftKey: boolean):boolean => {
+export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, shiftKey: boolean): boolean => {
 	const settings = getLatexSuiteConfig(view);
 
 	// Check whether we are inside a matrix / align / case environment
 	let isInsideAnEnv = false;
 
 	for (const envName of settings.matrixShortcutsEnvNames) {
-		const env = {openSymbol: "\\begin{" + envName + "}", closeSymbol: "\\end{" + envName + "}"};
+		const env = { openSymbol: "\\begin{" + envName + "}", closeSymbol: "\\end{" + envName + "}" };
 
 		isInsideAnEnv = ctx.isWithinEnvironment(ctx.pos, env);
 		if (isInsideAnEnv) break;
@@ -35,10 +35,10 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, 
 
 			setCursor(view, nextLine.to);
 		}
-		else if (shiftKey && ctx.mode.inlineMath) { 
+		else if (shiftKey && ctx.mode.inlineMath) {
 			tabout(view, ctx);
 		}
-		else if(ctx.mode.blockMath) {
+		else if (ctx.mode.blockMath) {
 			const d = view.state.doc;
 			const lineText = d.lineAt(ctx.pos).text;
 			const matchIndents = lineText.match(/^\s*/);


artisticat1:
Thank you for the PRs! I'll look into reviewing them soon.

src/features/matrix_shortcuts.ts
```diff
@@ -36,7 +36,15 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 			setCursor(view, nextLine.to);
 		}
 		else {
-			view.dispatch(view.state.replaceSelection(" \\\\\n"));
+			// Get the current line's indentation (indents), including all leading spaces and tabs.
+			const d = view.state.doc;
+			const lineNo = d.lineAt(ctx.pos).number;
+			const line = d.line(lineNo);
+			const lineText = line.text;
```
artisticat1, line 43: This can be simplified to `const lineText = d.lineAt(ctx.pos).text;`

src/features/matrix_shortcuts.ts
```diff
@@ -36,7 +36,15 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
...
+			const matchIndents = lineText.match(/^([	 ]*)/);
```
artisticat1, line 44: `lineText` only contains this line's text, so I don't think `^` is needed here. Can this be simplified to `/(\s*)/`?

Testudinidae: Great observation! I didn't notice that. Thanks!

Testudinidae:
I just rebased to version 1.9.8.

By the way, since I directly used:
```
git add .
```
the version in `package-lock.json` changed automatically. Should I revert it back?

src/features/matrix_shortcuts.ts
Testudinidae: As you wish.

Testudinidae:
For those who need complete auto-indent functionality, the snippet needs to be changed from this:
```typescript
{trigger: "pmat", replacement: "\\begin{pmatrix}\n$0\n\\end{pmatrix}", options: "MA"},
```
(1)

To this, ensuring that the \begin, \end, and the content in between are properly aligned:
```typescript
{flags: "m", trigger: "^(\\s*)(.*?)pmat", replacement: "[[0]][[1]]\\begin{pmatrix}\n[[0]]$0\n[[0]]\\end{pmatrix}", options: "rMA"},
```
(2)

If users need to add an additional level of indentation for the content, they might need to further modify it like this:
```typescript
{flags: "m", trigger: "^(\\s*)(.*?)pmat", replacement: "[[0]][[1]]\\begin{pmatrix}\n[[0]]\t$0\n[[0]]\\end{pmatrix}", options: "rMA"},
```
(3)

I think setting the default to the second format would be better. It doesn’t affect those who don’t need indentation, makes it more convenient for those who do, and for users who want an additional level of indentation, they can easily adjust it to the third format.

What do you think?

artisticat1:
> ... the version in `package-lock.json` changed automatically. Should I revert it back?

I think it's fine to leave it in.

> For those who need complete auto-indent functionality, the snippet needs to be changed from this:
>
> ...
>
> I think setting the default to the second format would be better. It doesn’t affect those who don’t need indentation, makes it more convenient for those who do, and for users who want an additional level of indentation, they can easily adjust it to the third format.
> 
> What do you think?

That makes sense. The only downside is that I find
```typescript
{flags: "m", trigger: "^(\\s*)(.*?)pmat", replacement: "[[0]][[1]]\\begin{pmatrix}\n[[0]]$0\n[[0]]\\end{pmatrix}", options: "rMA"},
```
is much less readable/intuitive than
```typescript
{trigger: "pmat", replacement: "\\begin{pmatrix}\n$0\n\\end{pmatrix}", options: "MA"},
```
— I think most users will find it easier to parse the latter.

As far as I'm aware, we haven't had any user requests for auto-indent functionality before. So to preserve the readability of the default snippets, what if we instead placed the auto-indenting snippets in the [snippets sharing thread](https://github.com/artisticat1/obsidian-latex-suite/discussions/50) for those who need them?

src/features/matrix_shortcuts.ts
```diff
@@ -38,9 +38,16 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 		else if (shiftKey && ctx.mode.inlineMath) { 
 			tabout(view, ctx);
 		}
+		else if(ctx.mode.blockMath) {
```
artisticat1, line 41:
```suggestion
		else if (ctx.mode.blockMath) {
```

artisticat1:
Thanks!

PR #383, opened 2025-01-29
[FEATURE] add matrix line spacing.

superle3:
Add the option for line spacing for the "Enter" matrix shortcut. For example `[1em]` in the setting becomes `\\[1em]` in the editor. 
Solves https://github.com/artisticat1/obsidian-latex-suite/discussions/381.

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -39,7 +39,8 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string, 
 			tabout(view, ctx);
 		}
 		else {
-			const lineBreakStr = (ctx.mode.inlineMath) ? " \\\\ " : " \\\\\n";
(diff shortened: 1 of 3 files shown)


artisticat1:
Thank you for taking the time to submit a PR!

However, I believe changing the default line spacing is better done via LaTeX/MathJax config, as opposed to appending `[1em]` to every single newline. (Alternatively, maybe [CSS like this](https://forum.obsidian.md/t/how-to-increase-space-between-math-lines/75643) could work.) For that reason, I'm closing this PR.

superle3:
This may be because I have no experience with css, but I believe that using css is more of a hassle for the user and a higher barrier of entry then a setting. For example if you only want it for specific files or specific parts of a file.

Also most importantly css would not transfer over with pandoc or to latex when you copy it over manually.

Although these use cases may be very niche for the time it could cost maintaining it. (I only needed it for my linear algebra assignments and haven't encountered it since.) 

Still thank you for reviewing this PR and responding so quickly.

Issue #378, opened 2025-01-20
[FEATURE] Option to disable snippets when typing arguments into curly brackets?


ofisurr:
# Description of the Problem
When trying to type the string in the curly brackets here as `\color{lightgreen}`, it detects `ee` as a snippet, making the output `\color{lightgre^{ n }}`.

# Description of the Solution
Perhaps an option where, whenever I'm typing in brackets, snippets are disabled all together.

![Image](https://github.com/user-attachments/assets/38b827b2-a38e-4cf8-b760-ee8ede065b98)

superle3:
Hello,
this has been fixed for the latest version and will be probably be added in the next release.

As for your suggestion, I don't think that it is usefull.
Most environments inside curly braces  need to be math since they are interpreted as math.
For example subscript, superscript and fractions. 

Currently a set amount of text environments are seen as text like `\text`.
`\color` was missing from this list and has now been added.

Still thank you for pointing out this bug.

EnriqueTMT:
A few other `\text` derivative environments that are not in the list yet. I have `\text` at the top as a "control" and everything below that have active snippets within the environment.

```
$$
\begin{align*}
    \text{Normal text} &: \text{This is normal text} \\
    \textbf{Bold text} &: \textbf{This is bold text} \\
    \textit{Italic text} &: \textit{This is italic text} \\
    \texttt{Typewriter text} &: \texttt{This is typewriter text} \\
    \textsf{Sans-serif text} &: \textsf{This is sans-serif text} \\
    \textrm{Serif text} &: \textrm{This is serif text}
\end{align*}
$$
```

They're: textbf, textit, texttt, textsf, and textrm. I just use textbf and textit, but I thought I'd list the others while I'm here.

superle3:
I have a pr open at https://github.com/artisticat1/obsidian-latex-suite/pull/385, but I am just not sure what to call a text environment. Technically \label is a math environment and `\label{\alpha}` is valid latex. Same goes for `\mathrm{\alpha}`, thats why its still a draft. Haven't a clear answer yet

superle3:
it works for these cases, just not for things with 2 arguments

PR #384, opened 2025-02-01
added mathrm and color to text environments inside math blocks

superle3:
solves https://github.com/artisticat1/obsidian-latex-suite/issues/378.

Ideally I would make this a setting since there are more text environments people may want to use like `\textit`, `\label` or something else but that don't make sense as default options.
But I first want to know if that feature is alright before I work on it.

### DIFF
--- a/src/utils/context.ts
+++ b/src/utils/context.ts
@@ -114,7 +114,9 @@ export class Context {
 			this.isWithinEnvironment(this.pos, {openSymbol: "\\text{", closeSymbol: "}"}) ||
 			this.isWithinEnvironment(this.pos, {openSymbol: "\\tag{", closeSymbol: "}"}) ||
 			this.isWithinEnvironment(this.pos, {openSymbol: "\\begin{", closeSymbol: "}"}) ||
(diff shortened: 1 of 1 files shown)


artisticat1:
Thanks!

> ... `\textit`, `\label` or something else but that don't make sense as default options.

Oh, I'd have thought they make sense to add as well - why not?

superle3:
I haven't used them, so I don't know if they make sense as default. But you also have `\textcolor` instead of `\color`.

 I  Use mostly `\text` for text and `\mathrm` for math things like `V_{\mathrm{Tot}}` or `\mathrm{Sinc}` . Besides that I don't think I use another text environment.
`\operatorname` would also make sense as a default.

But can it be a setting or not?

artisticat1:
Sure, we can make it a setting. We currently have a couple of settings for things which make better sense being hardcoded (e.g. the list of matrix environments), but are there in case a user wishes to tweak them.

> But you also have `\textcolor` instead of `\color` which look the same or `\textrm` instead of `\mathrm`. So adding every environment seemed unnecessary.

I think it makes sense to add all text environments one might conceivably use.

PR #385, opened 2025-02-02
Completed the list of text environments inside math environments. 

superle3:
This adds all the environments that don't accept math input in their arguments to the text environment in the context class.

Besides this another array was added.
It contains the macros where math input would be unusual but is still perfectly valid.
`\label{\alpha}` is for example correct syntax. (`\label` generally doesn't work great in mathjax/obsidian but thats unimportant)

> I think it makes sense to add all text environments one might conceivably use.

How should a text environment be defined. The mathfonts are generally treated as text like environments (from my experience),
but they are still math environments.

### DIFF
--- a/src/utils/context.ts
+++ b/src/utils/context.ts
@@ -5,6 +5,7 @@ import { Mode } from "../snippets/options";
 import { Environment } from "../snippets/environment";
 import { getLatexSuiteConfig } from "../snippets/codemirror/config";
 import { syntaxTree } from "@codemirror/language";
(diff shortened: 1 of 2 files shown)


superle3:
I think a setting where you can add your own text macros and math fonts is the best option. I can't think of a better compromise.

The setting itself could still use some css, but the implementation is done.

superle3:
I am gonna leave the user defined textareas as I am not sure how much freedom I want to give that setting. Like just an array of commands or more like is done for autofraction.

PR #391, opened 2025-02-13
Enhance matrixShortcuts: Optional whitespace trimming & multi-selection support

Testudinidae:
### Summary
This PR adds an optional whitespace trimming feature to `matrixShortcuts`.

### Changes
- **New setting**: `matrixShortcutsTrimWhitespace`
- **Functionality**:
  - When enabled, pressing Tab or Enter in a matrix environment will **remove surrounding whitespace** before inserting "&" or "\\\\".
  - **Multi-selection support**: Ensures consistent trimming across multiple cursors.
- **Purpose**: Prevents excessive spaces from affecting LaTeX matrix formatting.

This feature is **enabled by default** and can be disabled in settings.

### DIFF
--- a/src/features/matrix_shortcuts.ts
+++ b/src/features/matrix_shortcuts.ts
@@ -12,14 +12,31 @@ const LINE_BREAK_INLINE = " \\\\ "
 
 
 const generateSeparatorChange = (separator: string, view: EditorView, range: SelectionRange): { from: number, to: number, insert: string } => {
+	const [, separatorHeadSpace, separatorRemaining] = separator.match(/^([ \t]*)([\s\S]*?)$/);
+
 	const d = view.state.doc;
 
-	// Insert indents
-	const fromLineText = d.lineAt(range.from).text;
-	const leadingIndents = fromLineText.match(/^\s*/)[0];
-	const fixed_separator = separator.replaceAll("\n", `\n${leadingIndents}`);
+	const fromLine = d.lineAt(range.from);
+	const textBeforeFrom = d.sliceString(fromLine.from, range.from).trimStart();  // Preserve indents
 
-	return { from: range.from, to: range.to, insert: fixed_separator };
+	const toLine = d.lineAt(range.from);
+	const textAfterTo = d.sliceString(range.to, toLine.to);
+
+	let fixed_separator = "";
+	// If not at the beginning of the line
+	if (textBeforeFrom !== "") {
+		fixed_separator += separatorHeadSpace;
+	}
+	fixed_separator += separatorRemaining;
+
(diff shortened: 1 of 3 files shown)


src/features/matrix_shortcuts.ts
```diff
@@ -19,41 +74,42 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 
 	if (!isInsideAnEnv) return false;
 
-	// Take main cursor since ctx.mode takes the main cursor, weird behaviour is expected with multicursor because of this.
-	if (key === "Tab" && view.state.selection.main.empty) {
-		view.dispatch(view.state.replaceSelection(" & "));
+	trimWhitespace = settings.matrixShortcutsTrimWhitespace;
+
+	if (key === "Tab") {
```
superle3, line 79: matrix shortcuts shouldn't be run when text is selected. Indenting and deleting indentation is a more common use then replacing the selection with &.

src/features/matrix_shortcuts.ts
```diff
@@ -19,41 +74,42 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
...
+		applySeparator(ALIGNMENT, view);
```
superle3, line 80: shifting indentation when selecting something makes more sense then replacing the current selection with a &. Otherwise you can't change the indentation easily but you can replace the current selection by pressing & itself.

src/features/matrix_shortcuts.ts
```diff
@@ -19,41 +74,42 @@ export const runMatrixShortcuts = (view: EditorView, ctx: Context, key: string,
 
 	if (!isInsideAnEnv) return false;
 
-	// Take main cursor since ctx.mode takes the main cursor, weird behaviour is expected with multicursor because of this.
-	if (key === "Tab" && view.state.selection.main.empty) {
-		view.dispatch(view.state.replaceSelection(" & "));
```
superle3, line 24: shifting indentation when selecting something makes more sense then replacing the current selection with a &. Otherwise you can't change the indentation easily but you can replace the current selection by pressing & itself.


couldn't figure out how to review properly, sorry for the possible notifications

src/features/matrix_shortcuts.ts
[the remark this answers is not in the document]
Testudinidae: This is a good consideration. I will add `&& view.state.selection.main.empty` back, which means it’s based on the main selection. If there is a selection range in the main, it will indent; if there is no selection in the main, it will replace (regardless of other selections). But I’m also considering whether to change it so that if the selection is the entire line, it will indent, as some IDEs (like VS Code) handle it this way.

Testudinidae:
PR closed as it was replaced by another PR. #394

PR #394, opened 2025-02-19
Enhance Matrix Shortcut with Multi-selection

Testudinidae:
This is a significant update with multiple enhancements and refactors to improve matrix shortcut handling and internal logic. Below is a summary of the changes made:

- Matrix Shortcut - Multi-selection Support
Users can now use multi-selection within matrix shortcuts, enabling more efficient edits.
- Refactor `isWithinEnvironment`
The `isWithinEnvironment` function has been split into two separate functions:
  - `getEnvironmentBound`
  - `isWithinEnvironment`
This refactor improves clarity and enhances the function's usability.
- Handling `\hline`
The logic now ignores adding a line break (i.e., \\) after encountering \hline, ensuring no unnecessary line breaks are inserted in these situations.

Optional Features:
- Trim Excess Whitespace (default true)
When enabled, extra spaces around the cursor are trimmed automatically. This helps address common cases where users accidentally add extra spaces, either by pressing the spacebar unintentionally or due to snippets automatically adding them. For example, pressing <kbd>Tab</kbd> after `"1 "` will result in `"1 & "` rather than `"1  & "` to prevent accidental spaces from accumulating.
- Trim Excess Alignment (default true)
This trims unnecessary alignment symbols that can occur if users accidentally press <kbd>Tab</kbd> at the end of a line. For instance, pressing Enter after `"1 & "` could result in an extra `"&"` being inserted, but with this feature enabled, it will correctly turn into `"1 \\"` instead of `"1 & \\"` to avoid such mistakes.
- Enable `\\` After `\hline` (default false)
If enabled, a line break `\\` will be inserted after `\hline`.
- Trim Empty Line After Environment (default true)
When exiting an environment using <kbd>Shift+Enter</kbd>, an empty line after the environment is removed to prevent unnecessary blank spaces. For instance:
```LaTeX
...
    1 & 2 & 3 \\
     <- (cursor)
\end{pmatrix}
```
would transform to:
```LaTeX
...
    1 & 2 & 3 \\
\end{pmatrix} <- (cursor)
```
- Add Line Break After Environment (default false)
If enabled, Shift+Enter will add a line break after exiting an environment. For example:
```LaTeX
...
    1 & 2 & 3 <- (cursor)
\end{pmatrix}
```
will transform to:
````LaTeX
...
    1 & 2 & 3 \\
\end{pmatrix} <- (cursor)
```
These updates aim to enhance the user experience and provide better control over formatting and matrix shortcuts.


llakala:
Probably want to rebase and squash merge commits, as well as refactor commits.

src/features/matrix_shortcuts.ts
```diff
@@ -1,59 +1,178 @@
...
-		const env = { openSymbol: "\\begin{" + envName + "}", closeSymbol: "\\end{" + envName + "}" };
+		env = { openSymbol: "\\begin{" + envName + "}", closeSymbol: "\\end{" + envName + "}" };
 
 		isInsideAnEnv = ctx.isWithinEnvironment(ctx.pos, env);
 		if (isInsideAnEnv) break;
 	}
 
 	if (!isInsideAnEnv) return false;
 
-	// Take main cursor since ctx.mode takes the main cursor, weird behaviour is expected with multicursor because of this.
```
superle3, line 22: sorry to be nitpicky but why was this removed. Weird behavior  still arises when the main cursor is in different environment then the other cursors.
For example if nothing is selected in the main cursor but something is selected in another cursor.

Testudinidae: No need to apologize—I believe all features should be discussed to find the most convenient solution for everyone. I'm not sure what you mean by "weird behavior." I think in multi-selection cases, all selections should behave consistently—either all selected lines increase indentation or all selections are replaced with " & ", always following the main selection. Other potentially strange behaviors, like some selections being outside the environment or even outside the math block, shouldn’t be an issue because users using multi-selection should clearly know what they are doing. I hope you can provide more specific examples of what you mean by "weird behavior."

Testudinidae:
I might reorganize the git commit history in the next couple of days. Feel free to point out any unexpected behavior.

src/features/matrix_shortcuts.ts
superle3: My original thought was that a cursor according to its "environment" (like math, code or text).
 So any cursor that doesn't follow the behaviour of its environment is messy behaviour, but since I couldn't give a use case I just called it weird behaviour.
 Cause if you have the following code where | is the cursor
 
 ```tex
 \begin{pmatrix}
 1 & 2 |
 \end{pmatrix}
 | some selected text|

```
then either the lines are indented or the selected text is replaced with a &, which can be weird but also no use case.
The main cursor was indeed taken to be consistent with the rest of the plugin and put the comment there if someone wanted `ctx` cover multicursor.

But I still can't think of a use case, why someone would use cursors in different environments besides accidents, so it can be left out.

hope this explained it better.

File: src/features/matrix_shortcuts.ts [hunk 2]
```diff
@@ -1,59 +1,174 @@
...
+	const d = view.state.doc;
+
+	const fromLine = d.lineAt(range.from);
+	const textBeforeFrom = d.sliceString(fromLine.from, range.from).trimStart();  // Preserve indents
+
+	const toLine = d.lineAt(range.from);
+	const textAfterTo = d.sliceString(range.to, toLine.to);
+
+	let from = range.from;
+	let to = range.to;
```
superle3:
```suggestion
	let { from, to } = range;
```
this doesn't matter that much, but I prefer this style.

superle3 (commented):
will review further some time this week.

src/features/matrix_shortcuts.ts
```diff
@@ -1,59 +1,174 @@
...
+
+	if (trimWhitespace) {
+		// If at the beginning of the line
+		if (textBeforeFrom === "") {
+			separator = separator.match(/^[ \t]*([\s\S]*)$/)[1];
+		}
+
+		// Extend selection to include trailing whitespace before `from`
+		if (trimAlignment && isLineBreak(separator)) {
+			from -= textBeforeFrom.match(/\s*\&?\s*$/)[0].length;
```
superle3: useless escape, eslint warning. (I probl should introduce husky or smth to run the linter or introduce ci/cd)

src/features/matrix_shortcuts.ts
```diff
@@ -1,59 +1,174 @@
...
+
+const isLineBreak = (separator: string) => {
+	return separator.contains("\\\\");
+}
+
+
+const isMultiLineBreak = (separator: string): boolean => {
+	return separator.contains("\\\\") && separator.contains("\n");
+}
+
```
superle3: is there a specific reason why you're not checking for `LINE_BREAK` or just 
```ts
const regex = /\\\\\s*\n/
```
or smth? What custom separator did you have in mind? because there is no reason to have this generalization in the current version.

superle3:
I think thats all? Not  experienced with reviewing so if some comments are too outrageous or unclear, just let me know. 

(for example I am not sure how much I should govern the style, hopefully everything I said was reasonable?)

src/features/matrix_shortcuts.ts
```diff
@@ -1,59 +1,174 @@
...
+						applySeparator(END_LINE_BREAK, view);
+
+						d = view.state.doc;
+						nextLine = d.line(lineNo + 1);
+						newPos = nextLine.to;
+					}
+				}
+
+				setCursor(view, newPos);
+			}
```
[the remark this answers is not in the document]
Testudinidae: I’ll look into refactoring it so we can dispatch only once where possible.

superle3:
Hi @Testudinidae ,

(Not trying to rush you, it's oss and volunteer work so things take time/ take your time)

I am not in a hurry/dont have a lot of free time this week, but do you mind me merging other prs and make a release since this pr is in draft or were you planning on completing it soon?

Also since you resolved every comment in the other pr (except the last one but that doesn't need to be answered before merging), does that mean I can merge it?

PR #398, opened 2025-02-27
Bump version to 1.9.9 and optimize performance of cursor movement in LaTeX blocks

HerrChaos:


src/latex_suite.ts
```diff
@@ -29,82 +29,83 @@ export const handleUpdate = (update: ViewUpdate) => {
...
-	if (settings.matrixShortcutsEnabled && ctx.mode.strictlyInMath()) {
-		if (["Tab", "Enter"].contains(key)) {
-			success = runMatrixShortcuts(view, ctx, key, shiftKey);
-
-			if (success) return true;
+		
+		// Matrix shortcuts
+		if (settings.matrixShortcutsEnabled && ["Tab", "Enter"].contains(key)) {
+			if (runMatrixShortcuts(view, ctx, key, shiftKey)) return true;
 		}
```
Testudinidae, line 99: Since the <kbd>Tab</kbd> key event has already been intercepted earlier, it never reaches this point when <kbd>Tab</kbd> be pressed.

Testudinidae (commented):
Thanks for your work on this! However, the original logic prioritizes `runMatrixShortcuts` over `tabout`, so this might change the intended behavior. What do you think?

HerrChaos:
Hello, thanks for looking into my code. The bit you commented on was an accidental leftover from testing, since I was not entirely sure what cause the lag at the start. I am sorry for wasting you time with this. Thanks again for the review. I have updated the code to just revert that section to it's original form.

HerrChaos:
Also I just want to add for future reference. I think the main lag is coming from the little preview renderer which pops up at the top or bottom.

superle3:
If you'll also remove the version bump and the `requestAnimationFrame`, then I'll merge it.

HerrChaos:
> If you'll also remove the version bump and the `requestAnimationFrame`, then I'll merge it.

Done, thank you very much for your time <3

Issue #335, opened 2024-09-29
[BUG]  The plugin doesn't work on Android, as shown in the video


CamWam:
https://github.com/user-attachments/assets/ea6c3d26-4795-4e0a-ad62-8995b75ca3c5

As shown in the video above, restarting Obsidian, reinstalling the plugin, and deleting the script directory did not resolve the issue. It appears that the script may not work on Android.

melody15383800870-design:
换一个输入法
change a input method

superle3:
In the next release (2-4 weeks) the fix will be added.

Issue #326, opened 2024-09-13
[BUG] Plugin does not work on Android


dwd-1123:
# Description
The plugin (also includes Quick Latex for Obsidian) won't work properly on Huawei tablets and Android emulators (Mumu Player 12), but it will work on Windows 10 22H2 and iOS 16.4 (IME: Gboard)

# Expected Behavior
The plugin should work properly on A ndroid devices

# Additional Info
Only Conceal part can work on Android

# Platform
Huawei tablet:
Huawei Matapad Pro 10.8/MRX-W09
HarmonyOS 4.0.0 HMS Core 6.14.0.302
IME: Gboard
Obsidian version 1.6.7(149) API v1.6.6

Android emulators:
Mumu Player 12 Version:V3.8.18(2845)
Android 12
External physical keyboard
Obsidian version 1.6.7(149) API v1.6.6

superle3:
in the next release (2-4 weeks) the fix will be added.

Issue #213, opened 2023-10-19
[BUG] Plugin not working on Samsung tablets after upgrading to OneUI 5.1.1


lukks22:
# Description
Since the upgrade, me and (at least) another user can't use the shortcuts anymore, not even mk, dm or any custom ones.

# Expected Behavior
Shortcuts should have effect

# Additional Info
None

 # Platform
Samsung tablets (found on Samsung Galaxy S7 FE and Galaxy Tab S6 Lite) with One UI 5.1.1

Before the update the shortcuts would not work when using the "virtual" keyboard, but would work with a bluetooth one. After the update not even that. The same project on Windows works fine.

artisticat1:
Can you try out the build in #191 and check whether the plugin works again?

lukks22:
> Can you try out the build in #191 and check whether the plugin works again?

I'm not very familiar with GitHub, if you could tell me how I'd be more than happy to

artisticat1:
> I'm not very familiar with GitHub, if you could tell me how I'd be more than happy to

Of course. Download the `main.js` file from https://github.com/oldkingOK/obsidian-latex-suite/releases/tag/v0.0.4 (which corresponds to the fix in #191 by @oldkingOK).

Then replace `path-to-vault/.obsidian/plugins/obsidian-latex-suite/main.js` with the new `main.js`.

> Just update the plugin in obsidian and check if it works for you.

The fix in #191 actually hasn't been released as an update yet, so if you could follow the steps above and report back that'd be great!

lukks22:
Sorry man, still not working. Some snippets work just fine (a/b for frac, << for /ll) but not mk, dm or any other.

I've tried both on the physical keyboard and the system's one

oldkingOK:
Hi @lukks22 .
I don't have any Samsung machines around, so it would be helpful if you could describe the situation in more detail.
- What keyboard app are you using?
- Is the composition mode off? (when on, words are underlined during spelling). You can view [how to turn off it for some keyboards here](https://github.com/oldkingOK/obsidian-latex-suite/discussions/4).

oldkingOK:
> The on-screen keyboard doesn't work with text prediction on but the physical one does.

I'm having trouble dealing with the text prediction function.

1. Pressing TAB during spelling has a high probability of causing the cursor to [lose focus](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event).
[...]

PR #191, opened 2023-09-16
fix: support android app

oldkingOK:
See [#52](https://github.com/artisticat1/obsidian-latex-suite/issues/52#issuecomment-1722239752)

PR #403, opened 2025-03-05
Fix Android by force end composition

### LINKED ISSUE #52: Mobile support does not seem to work
The support for mobile (android) doesn't seem to work. 

Recording

I tried it on both of my android devices but neither of them worked. I checked if other plug-ins caused that, so I made a fresh obsidian installation but it still has no effect.

**Additional Details:**
- Live Preview editor is enabled
- Lastest version is installed
- Android version: 12

oldkingOK:
Resolves #52 

Due to the presence of IME, this plugin cannot run on most mobile devices. Although this PR does not adapt to all mobile devices, based on user feedback in the issue, it has at least successfully worked on some devices.

These patches only work when `Don't trigger snippets when IME is active` is turned off. I think we should provide a prompt for all mobile users, but I am not familiar with this part of the code, so I need some help.

Most of the code in this patch is based on the vim-CM6 plugin, and I am a beginner in plugin development, so there may be many issues in the code. Any suggestions are very welcome!

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -52,6 +52,7 @@ export default class LatexSuitePlugin extends Plugin {
 			return;
 		}
 	}
+
 	IMEEditorWarning() {
 		if (isIMESupported() && !this.settings.suppressIMEWarning) {
 			// const message = "Obsidian Latex Suite: this plugin supports your IME keyboard, but `Advanced settings > Don't trigger snippets when IME is active` is currently disabled. You may want to enable it in the plugin settings to avoid unexpected snippet triggers while using your IME. To turn this warning off, enable `Advanced settings > Suppress IME warning`.";
@@ -64,7 +65,6 @@ export default class LatexSuitePlugin extends Plugin {
 			new Notice(message, 10000);
 			console.log(message);
 		}
-		
 	}
 
 	async loadSettings() {
--- a/src/settings/settings_tab.ts
+++ b/src/settings/settings_tab.ts
@@ -469,10 +469,11 @@ export class LatexSuiteSettingTab extends PluginSettingTab {
 				this.plugin.settings.autoDelete$ = value;
 				await this.plugin.saveSettings();
 			}));
-		const surpressIMESetting = new Setting(containerEl)
+
+		const suppressIMESetting = new Setting(containerEl)
 			.setName("Don't trigger snippets when IME is active")
 			.setDesc("Whether to suppress snippets triggering when an IME is active.");
-		const surpressIMEWarning = new Setting(containerEl)
+		const suppressIMEWarning = new Setting(containerEl)
 			.setName("Suppress IME warning")
 			.setDesc("Whether a warning is shown on startup if `Don't trigger snippets when IME is active` is enabled. Disable this if you are aware of the IME limitations. Currently only ios and android have support for IME")
 			.addToggle((toggle) => toggle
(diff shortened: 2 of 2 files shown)


superle3:
Should also fix #335, #326 and #213 right?

superle3:
What do you mean, with a prompt?

Do you want to prompt them on install or on every launch?
It could also be turned off by default for android?

It should also be mentioned either in the docs or in the settings, or maybe a combination of both?

superle3:
Hi @oldkingOK ,

Since it's just a style change, I'll rebase and merge this in about 2 weeks if I don't hear from you before that.

oldkingOK:
Hi @superle3 ,

Sorry for the late reply, finally got some time to get back to this.

> What do you mean, with a prompt?
> Do you want to prompt them on install or on every launch?
> It could also be turned off by default for android?

Yes, that’s what I meant. My initial plan was to show a prompt once right after installation. However, turning this off by default on Android and iOS sounds like a great idea, since the option was originally designed for Windows and doesn't provide any benefit on mobile devices.

> Should also fix #335, #326 and #213 right?

I don’t have the devices referenced in those issues, but I tested the same input on my Xiaomi Pad 5 and it works fine. This PR could serve as a good starting point for Mobile compatibility, and we can adjust further based on user feedback.

oldkingOK:
However, adding separate logic for iOS and Android would introduce some platform-specific code. As artisticat1 mentioned [here](https://github.com/artisticat1/obsidian-latex-suite/pull/191#discussion_r1334982322):

> The goal is to make Latex Suite a pure CodeMirror extension, with no dependency on Obsidian.

So I'm not entirely sure how to handle this part cleanly, and I'd appreciate further guidance or suggestions.

superle3:
> Sorry for the late reply, finally got some time to get back to this.

Its fine, thanks for replying so quickly  (not sarcastic, don't know how to phrase this otherwise)

> However, adding separate logic for iOS and Android would introduce some platform-specific code. As artisticat1 mentioned https://github.com/artisticat1/obsidian-latex-suite/pull/191#discussion_r1334982322:
>>The goal is to make Latex Suite a pure CodeMirror extension, with no dependency on Obsidian.

well, its not obsidian specific, atleast on electron the wrapper around macos, windows and linux is just that a wrapper, but they don't use electron for mobile but capacitor (probl smaller size), but that still has workarounds to detect the user agent, see https://stackoverflow.com/questions/21741841/detecting-ios-android-operating-system for reference.

But since we already rely on obsidian to create the settings_tab, its fine to use `Platform` in there. 
> So I'm not entirely sure how to handle this part cleanly, and I'd appreciate further guidance or suggestions.

since obsidian doesn't have a warning system like vscode, the best option is to create another setting that turns the warning on and off when `Don't trigger snippets when IME is active` is on. 
By default the warning should be on and it should at least explain how to turn it off, ~~but i don't know if this should be done solely for android/ios or also for other platforms.~~
Only android/ios, as only android's/ios "IME" keyboard is "supported", we can add the warning later for other platforms when that is supported, but now it would be useless. 

having it default to off would not help the users that have it currently installed, instead a section in the readme needs to warn IME/ios/android users and point to another section in the `DOCS.md`.


> However, turning this off by default on Android and iOS sounds like a great idea, since the option was originally designed for Windows and doesn't provide any benefit on mobile devices.

I don't think you can do this cleanly, while `settings_tab.ts` is connected to obsidian, `settings.ts` is not. I think the warning and a section in the readme is enough.

superle3:
Something like https://github.com/artisticat1/obsidian-latex-suite/pull/403/commits/be140d851e81ffdbc4d4cad20977200ebd887757 (you can force push/rebase with git if you want to delete that commit)

oldkingOK:
> but that still has workarounds to detect the user agent

Checking the device just by looking at the user agent feels a bit too hacky. Since we can already use Obsidian’s API in the `settings_tab`, I think it's totally fine to rely on that for platform detection.

> having it default to off would not help the users that have it currently installed

Yeah, you're right — I didn't think about that before.

> Something like be140d8 (you can force push/rebase with git if you want to delete that commit)

Thanks for the reference! That already gets the job done. I just fixed some tiny typos.

superle3:
> Checking the device just by looking at the user agent feels a bit too hacky. Since we can already use Obsidian’s API in the settings_tab, I think it's totally fine to rely on that for platform detection.

I meant more that you would be able to "mock" this behavior in the web and that its not "really" obsidian specific, and you can spoof the user-agent but codemirror itself thinks the user-agent is good enough see https://github.com/codemirror/view/blob/076450d0bee332d3eee1cd72105c31add7f83de3/src/browser.ts

PR is complete so I am merging now, I will still wait for @Testudinidae  to respond for 2-3 weeks, and then I'll see about creating a release and merging my prs.

superle3:
Fix is out in 1.10.0, sorry for the delay

PR #412, opened 2025-03-26
[FEATURE] Disable Auto subscript when in \ce environment

### LINKED ISSUE #411: [FEATURE] Disable Auto subscript when in \ce environment
Is there a way to disable auto subscrip when the cursor is in \ce? 

Because by default when I input a _number_ after a _letter_, it turns it into `letter_{number}`. But ce does a better job handling `letternumber`. 

E.g., if I do `P4O10`, 10 will all be subscripped. But when latex suite changes it to `P_{4}O_{1}0`, it is less readable when unrendered, and 0 is in normal size.

superle3:
Fixes #411

### DIFF
--- a/src/snippets/environment.ts
+++ b/src/snippets/environment.ts
@@ -9,8 +9,10 @@ export interface Environment {
 /**
  * a mapping of triggers to environments where they should not run
  */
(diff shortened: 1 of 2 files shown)


artisticat1:
Thanks!

PR #418, opened 2025-04-07
Feature "Symbol groups", and more changes.

alexanderNB:
This adds a new setting called "Symbol groups", Here you can, much like how snippet variables work, create different symbol groups which is parsed into Snippet Variables and another new setting that I will talk about in a bit.
As an example if you in the symbol groups setting have:
{
    "{GREEK}": "alpha|beta|gamma"
}

And then in the snippet variables setting have:

{
	"${GREEK}": "{GREEK}|Gamma|delta|..."
	"${OTHER_SNIPPET_VARIABLE}": "{GREEK}|sin|cos"
}

Then that would be the exact same as writing:
{
	"${GREEK}": "alpha|beta|gamma|Gamma|delta|..."
	"${OTHER_SNIPPET_VARIABLE}": "alpha|beta|gamma|sin|cos"
}

This allows for more detailed descriptions of snippet variables, which might even only be used at one location in the actual snippets.

I have made changes to default_snippets.js, default_snippet_variables.js and added default_symbol_groups.js to showcase how this new system improves upon the old one.

Before I had issues with the old default snippet variables, as they where in no way descriptive of what they actually did, and maybe you would want a specific symbol in one of the cases where it uses "${SYMBOL}" but not in the other. But I could also see how in the old system it would have been really cluttered to split them. The ${GREEK}: "{GREEK}" default snippet variable I have mainly kept to show that it's allowed to call a snippet variable value "the same" as the symbol groups. 

Another reason why I wanted to build this system is that it annoyed me, to a reasonable degree, that in the autofraction.ts code there was a system that made greek symbols also come under the fractions. E.g
"\\alpha a/b" -> "\\frac{\\alpha a}{g}" which is a great system, but it was hardcoded to only greek symbols, so a new setting I have made called "Included symbols", now allows you to define which symbols this would work for. This setting will also use symbol groups
By default I have set it to:
{GREEK}|{SYMBOL}
Which means that all of the symbols in those two symbol groups will allow for this same functionality.

I have also made a few other changes such making the snippet variable also work in the `replacement` of a snippet, but only if it's a function, by default this is used in a snippet which takes care of auto letter subscript.

The UI of the snippet variables, and symbol groups, settings now look better, it's close to how the snippets editor looks, with a few minor changes. Under the snippet variables editor, I have also included a preview button, which allows you to see how the snippet variables actually look, after symbol groups have been inserted.


superle3:
Probably should mention it in the docs and not readme. Snippet variables make writing snippets easier but it doesn't affect how you write latex and is more an advanced feature.

superle3:
Hi, 
Thank you for making this PR.

I'll try to review this pr and the other open prs as soon as possible, but it may take me some time due to the amount.

Though before I review I would like you to clarify some things.

> I have also made a few other changes such making the snippet variable also work in the replacement of a snippet, but only if it's a function, by default this is used in a snippet which takes care of auto letter subscript.

Why is this a macro instead of passing it onto the function? The macro could cause a parsing issue (though probably not likely) and it's also unnecessarily complex.


> This adds a new setting called "Symbol groups", Here you can, much like how snippet variables work, create different symbol groups which is parsed into Snippet Variables and another new setting that I will talk about in a bit.

It's that I'm more programmer oriented than user so this might be biased, but isn't this better handled by using JavaScript variables? If one uses an external file they can structure the code however they want and it doesn't add extra complexity. (You can also do it in the editor with an iffe, but at that point the syntax highlighting of an editor is preferable).

superle3:
With #568 I don't believe this would be the right direction anymore and more inclined to handle it by js instead.

Issue #413, opened 2025-03-27
[FEATURE] Minor improvement: Insert node in numerator on empty `\frac`


haasal:
# Description of the Problem
When typing `()/` the completion is `\frac{}{|}` where `|` is the cursor position. This doesn't fully make sense.
I would expect that this expands to `\frac{|}{}` -> `TAB` -> `\frac{}{|}`.

# Description of the Solution
I haven't looked at your code yet but the solution is probably 3-5 lines of code *max*. Basically check if the numerator capture group is empty and if yes return a snippet with two "insert-nodes" instead of one.

# Additional
...

# Additional context
In luasnip I implemented this as follows (`mzs(...)` is my own function so you wont find a signature like this in the luasnip docs but I think the point is clear).

```lua
  mzs(
    [[\((.*)\)\/]],
    d(1, function(_, parent)
      local numerator = parent.captures[1]
      if #numerator == 0 then
        return sn(nil, fmta("\\frac{<>}{<>}", { i(1), i(2) }))
      else
        return sn(nil, fmta("\\frac{<>}{<>}", { t(parent.captures[1]), i(1) }))
      end
    end),
    { trigEngine = "ecma" }
  )
```

**I will happily make a PR once this is approved!**

PR #436, opened 2025-07-16
feat: implement #413,

superle3:
()/ -> \frac{$0}{$1}$2 makes more sense then () -> \frac{}{$0}$ so the first one is chosen when there is nothing inside the brackets.

Using `//` would be better, but it may be easier/more natural to think of `()/` instead of `//`.

### DIFF
--- a/src/features/autofraction.ts
+++ b/src/features/autofraction.ts
@@ -106,9 +106,15 @@ export const runAutoFractionCursor = (view: EditorView, ctx: Context, range: Sel
 		}
 	}
 
-	const replacement = `${settings.autofractionSymbol}{${numerator}}{$0}$1`
+	// If the content inside parentheses is empty, the numerator would be empty and that's rarely desired.
+	const replacement =
+		numerator === ""
+			? `${settings.autofractionSymbol}{$0}{$1}$2`
+			: `${settings.autofractionSymbol}{${numerator}}{$0}$1`;
+	// The keypressed shouldn't be inserted back in after an undo, if we have a selection.
+	const keyPressed = from != to ? undefined : "/";
 
-	queueSnippet(view, start, to, replacement, "/");
+	queueSnippet(view, start, to, replacement, keyPressed);
 
 	return true;
 }


src/features/autofraction.ts
```diff
@@ -105,8 +105,13 @@ export const runAutoFractionCursor = (view: EditorView, ctx: Context, range: Sel
...
 	}
-
-	const replacement = `${settings.autofractionSymbol}{${numerator}}{$0}$1`
+	
+	let replacement: string;
+	if (numerator === "()") {
+		replacement = `${settings.autofractionSymbol}{$0}{$1}$2`;
+	} else {
+		replacement = `${settings.autofractionSymbol}{${numerator}}{$0}$1`;
+	}
```
artisticat1, line 114: The ternary operator can be used here for conciseness:

```suggestion
	
	const replacement = numerator === "()" ? `${settings.autofractionSymbol}{$0}{$1}$2` : `${settings.autofractionSymbol}{${numerator}}{$0}$1`;
```

superle3: Thought about that and the snippet below
```ts
numerator = numerator || "$0"
const replacement = `${settings.autofractionSymbol}{${numerator}}{$1}$2
```
but wasn't sure if I wasn't "complicating" things further.

numerator should be checked against an empty string instead of `()`. 
I somehow didn't catch that when I went from testing my code to finalizing it

superle3:
While implementing  #443, I noticed that the parameter `keyPressed` is a string and not `undefined` regardless if  its automatic, visual, or using the snippetTrigger. Is this intentional?
For autofraction the current behaviour is this:

https://github.com/user-attachments/assets/30fb79a0-de7b-4378-95c1-51efc649b65c

instead of this:


https://github.com/user-attachments/assets/69f999b2-4c0a-40cf-b7fc-1b2632941660

The difference is, the key pressed that caused the visual snippet gets appended to the selection and two undo commands instead of one need to be executed to go back to before the visual snippet

Issue #439, opened 2025-08-12
[FEATURE]


FRIEDparrot:
# Description of the Problem
Insert math sign by hotkeys override when in math environment. 

# Description of the Solution 
In my habit to write equations, I use `Templater` plugin and configure a lot of hotkeys to insert math symbols, which is even faster than write a short-hand of the symbol.  For example, I use the following hotkeys : 
```py
alt+d -> \delta 
shift+alt+d -> Delta 
ctrl + alt + d -> \dot 
shift + ctrl + alt + d -> \ddot  
alt + w -> \omega  
alt + shift + w -> \Omega  
alt + e -> \epsilon 
alt + shift + e -> \varepsilon  
alt + x | c | v -> \xi | \cdot  |  \times 
alt + shift + x| c | v  ->   \sin | \cos | tan   
```
For example, write `omg` may be much slower than just `alt+w`. So I think it's good for this plugin to provide an option to override the hotkeys in math environment and use a script to customize what to insert. 

# Additional 
sometimes after inserting a math symbol, We need to put the cursor into a new location. 
For example, I use alt + ` to insert `\tilde{}`, in that case the script should be able to specify where the cursor is (make it into the bracket) in the script. 

I think this will be a good feature for this plugin, and it helps to improve math-writing efficiency and ergonomics😄.

Very appreciate it if u can take this into consideration!

FRIEDparrot:
Hi! Thanks for implementing this new feature!
Just asking for how to use this feature now.
the idea to add a "triggerKey" for it is really good. 

But now if I use this 
```json
{
      trigger: "d", 
      replacement: "\\delta", 
      options: "m", 
      triggerKey: "Ctrl-d"
  }
```
It triggers after typing `d` and Tab (not Ctrl-d).
For this feature I think when triggerKey is pressed, we just need to search and match the triggerKey field. 
- Press Alt-d directly
- Gets \delta immediately (no typing required) 

But now `Ctrl-d` seems not work now but `type d+ press tab` works. 
So did the triggerKey parsed correctly? 
Also I think we needn't check the trigger If triggerKey already matched. 
Thanks for your considerations and such nice works!

superle3:
its not yet released? or did you built with that pr included because tab doesn't work on my end. Also c-d is a vim keymap so there might be conflict there.

> Also I think we needn't check the trigger If triggerKey already matched.

Why? that doesn't make sense to me. The trigger can be empty so
```ts
const snippet = {
[...]

Issue #400, opened 2025-02-28
[FEATURE] Customizable Hotkey for "Tabout"


ankitp94:
# Description of the Problem

Is it possible to assign other hotkey, such as `Ctrl + Space`, for triggering tabout (instead of the usual Tab key)?

  I am currently using Tab for triggering certain snippets. Therefore, when I press a tab, it often triggers a matching  snippet instead of the intended Tabout.

# Description of the Solution

An option to select a hotkey for triggering Tabout action.

Sorry if I missed this option in the documentation. Thanks again for designing such a great extension!

superle3:
Hello,

the option doesn't exist yet.
Would a regular obsidian shorcut suffice and should this also disable tabout with the tab key.

ankitp94:
Yes, a regular obsidian shortcut should suffice and yes it should disable the tab key for tabout. Thank you!

PR #443, opened 2025-08-28
[Feature] customize keymaps

superle3:
Rely on @codemirror/view for handling keymaps instead rolling our own. This allows for more customization later on.

This tries to implement #439 #400.

I looked at the implementation of https://github.com/codemirror/view/blob/main/src/keymap.ts to see if it was worth it to roll our own keymap handling, but it was a bit too complex to include it, thus this relies on their implementation and adds the scope `latex-suite` such this extension only triggers it.


src/settings/settings_tab.ts
```diff
@@ -342,6 +340,17 @@ export class LatexSuiteSettingTab extends PluginSettingTab {
 					this.plugin.settings.taboutEnabled = value;
 					await this.plugin.saveSettings();
 				}));
+		new Setting(containerEl)
+			.setName("Key trigger for tabout")
+			.setDesc("What key to press to trigger tabout. Key names may be strings like \"Shift-Ctrl-Enter\" a key prefixed with zero or more modifiers. Modifiers can be given in any order. Shift- (or s-), Alt- (or a-), Ctrl- (or c- or Control-),  Cmd- (or m- or Meta-) or Mod- (Cmd- on Mac, Ctrl- on Windows/Linux) are recognized. Chords are seperated by a space.")
```
superle3, line 345: Am not really satisfied with this explanation

src/snippets/parse.ts
```diff
@@ -108,6 +109,7 @@ const RawSnippetSchema = object({
 	flags: optional(string_()),
 	priority: optional(number()),
 	description: optional(string_()),
+	triggerKey: optional(string_()),
```
superle3, line 112: following the same style here, but defaultvalue and validator could also be used.

PR #433, opened 2025-07-05
fix: don't see tags with `math` as an equation

superle3:
In obsidian, the name of a tag is the following hashtag_hashtag-end_meta_tag-<tagname>, if this tag name were to contain the word `math`, then it would recognize the whole tag as an equation instead of a tag. Fix is to check if we are not currently in a tag. fixes #431.

PR #445, opened 2025-09-14
[Feature] optimize conceal

superle3:
Optimize the scrolling when conceal is enabled.
Currently the whole concealment array is being recomputed even though nothing has been changed.
The change is not that much from 4-8 ms to <1ms on my machine, but the little things may add up.

There is also currently a bug, where the syntaxtree hasn't been fully computed yet.
To reproduce: Reload page without saving and go bottom with `G` in vim.
`ensureSyntaxTree(state,to)` can be used or we can debounce a view.dispatch when its not available.
@llakala do you maybe have any comment? `ensureSyntaxTree` didn't take more than 30ms on my machine even with 22,222 words,  but `ensureSyntaxTree` only updates a local tree and not the global syntax tree.

Also the bug is very minor so idk if its worth the extra complexity.

### DIFF
--- a/src/editor_extensions/conceal.ts
+++ b/src/editor_extensions/conceal.ts
@@ -252,13 +252,18 @@ export const mkConcealPlugin = (revealTimeout: number) => ViewPlugin.fromClass(c
 	decorations: DecorationSet;
 	atomicRanges: RangeSet<RangeValue>;
 	delayEnabled: boolean;
+	concealSpecs: ConcealSpec[];
+	lastVisibleRange: readonly { from: number, to: number }[] = [];
+	performance_log = {average: 0,weight: 0}
 
 
-	constructor() {
+	constructor(view: EditorView) {
 		this.concealments = [];
 		this.decorations = Decoration.none;
 		this.atomicRanges = RangeSet.empty;
 		this.delayEnabled = revealTimeout > 0;
+		this.concealSpecs = null;
+
 	}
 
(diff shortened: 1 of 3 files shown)


llakala:
I've been using Neovim with Typst for notes recently, so haven't been touching Obsidian!

Concept looks good to me. Honestly, I think it may be worth it to commit to a minimal fork. If @artisticat1 becomes more active again, you can always upstream individual commits back in - but you have a lot of open PRs that haven't received review, and I think you deserve a place to consolidate these patches for anyone to use.

superle3:
Thank you for responding so quickly,
Am currently in the process of porting this plugin to overleaf/ am extracting all the obsidian components out of this plugin at [here](https://github.com/superle3/snippet-leaf) and the performance issue was worse on overleaf.
So I first tried to fix it on obsidian since it has a better dev environment and got into that bug.
But you will only notice that bug when using vim, so was curious if you noticed it (and thus if it was worth fixing).

> Honestly, I think it may be worth it to commit to a minimal fork.

In #433 @artisticat1  said
> I graduated and don't write much LaTeX anymore, and also don't have as much time to review PRs as I used to - would you be interested in me adding you as a collaborator on this repo? That way, you (and other contributors) wouldn't be held up by me no longer having the time to devote to this project :)

I said yes? but no response?, not sure what is going on there. 
If the silence is very long and I have incorporated everything I want in the overleaf port, I may base a fork on [the npm package](https://www.npmjs.com/package/codemirror-latex-suite) instead of this.

PR #456, opened 2025-10-13
remove conflict with delta greek symbol

wuacs:
Using the default settings one can't write the delta symbol writing "delta" as this outcompletes to \nabla(i don't know why).
Since we have already the "nabl" trigger I think it is safe to delete the "del" trigger which causes conflicts with the "delta" symbol.


superle3:
> Using the default settings one can't write the delta symbol writing "delta" as this outcompletes to \nabla(i don't know why).


https://en.wikipedia.org/wiki/Del, its called the del or nabla operator. Naming things is hard.

> Since we have already the "nabl" trigger I think it is safe to delete the "del" trigger which causes conflicts with the "delta" symbol.

Agreed, but since it was added can you comment it out instead and have small explanation above it?

wuacs:
Done

PR #509, opened 2026-02-27
wip: conceal fixes

### LINKED ISSUE #182: [BUG] Mouse selection highlights wrong text
Write the following inline equation in Obsidian (this is just an example):

```
$\epsilon_{i} = \zeta_{+}U_{i}^{+} + \zeta_{-}U_{i}^{-}$
```

Now try to select this equation, by clicking on the right end (just before the ending `$`) and dragging the mouse (with the click held) to the left, until the beginning (just after the first `$`). Do not include the `$`'s in the selection. Then, you will see the following gets highlighted as the selection:

Screenshot 2023-08-26 at 01 36 46

which is wrong of course because we wanted to select the full equation, not just this portion. Apparently the concealment confuses the selection highlight logic. In fact, in reality the full equation is selected, as you can check by doing Ctrl+C and Ctrl+V somewhere, the full equation gets pasted.

So the bug is only visual, the selected "highlighted" text does not correspond to the actual selected text. I think it's related to how concealment expands those greek letters when selected.

superle3:
atttempt to fix #182

### DIFF
--- a/src/editor_extensions/conceal.ts
+++ b/src/editor_extensions/conceal.ts
@@ -246,6 +246,15 @@ function buildAtomicRanges(concealments: Concealment[]) {
 	return builder.finish();
 }
 
+const updateSelection = debounce((view: EditorView) => {
+	const ranges = view.state.selection.ranges.map(r => {
+		return EditorSelection.range(r.from, r.to, r.assoc);
+	});
+	view.dispatch({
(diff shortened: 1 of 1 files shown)


superle3 (commented):
still not really ideal cause there is some delay which has to be there but better than nothing

PR #510, opened 2026-02-27
wip: fix nested inline math

### LINKED ISSUE #153: [BUG] Math mode snippets trigger inside \text{}
# Description
While writing text inside `\text{}`, snippets with the `"m"` option are triggered

# To Reproduce
`$$\text{text with math inside $\sum$ }$$`
writing something like `gg` inside `\text` but outside `$$` will trigger the snippet (`\gg` in this case)

# Expected Behavior
The snippet should not trigger if it has the `"m"` option set

# Additional Info

https://github.com/artisticat1/obsidian-latex-suite/assets/42670032/068ba7dc-7328-4c3b-856d-1d466e16aa27

superle3:


superle3:
closing due to #581

PR #511, opened 2026-03-01
added snippet for displayed math inside lists

Sinthoras7:
Before, when I pressed `1. Test dm` this expanded to
```
1. Test $$
<cursor>
$$
```
I find this annoying, because most of the time I want to stay inside the list item even with displayed math. So this snippet changed is, such that now this is expanded to
```
1. Test $$
   <cursor>
   $$
```
This also works with normal bullet lists and on deeper layers.

As a disclaimer, the snipped was written by Chatgpt and I don't understand it fully.

### DIFF
--- a/src/features/tabout.ts
+++ b/src/features/tabout.ts
@@ -150,9 +150,9 @@ export const tabout = (view: EditorView, ctx: Context): boolean => {
 
 		// Trim whitespace at beginning / end of equation
 		const currentLine = doc.lineAt(cursorPos);
-		if (currentLine.text.trim() !== currentLine.text) {
+		if (currentLine.text.trimEnd() !== currentLine.text) {
 			transactions.push({
-				changes: {from: currentLine.from, to: currentLine.to, insert: currentLine.text.trim()}
+				changes: {from: currentLine.from, to: currentLine.to, insert: currentLine.text.trimEnd()}
 			})
 		}
 		view.dispatch(...transactions);


superle3:
ehh, you can't really parse markdown with regex, would be better if you had access to the syntaxtree through the function.
So like
```ts
type  replacement = (match: ..., {state: EditorState, syntaxtree: ..., })
```
and check if the current line is the beginning of a list.
But I am not sure rn about the api or form of that, I'll look at it later.

> As a disclaimer, the snipped was written by Chatgpt and I don't understand it fully.

Thank you for being honest, but you do understand that I as the reviewer need to understand the code and possibly ask questions about it?
Like why priority 50 and 100 was used, that seems excessive.
`^([ \t]*)(\d+[.)]|[-*+])(\s+)(.*?)\bdm\b$/m`
multiline is disabled by latex suite.
~~it allows 1) but only 1. is markdown syntax and~~ \s+ allows newlines which is not needed anymore in 1.11 where the same indentation level is kept.

Besides displaymath in readingview/ markdown spec needs to start on a newline if it spans multiple lines so the only either you press `Shift-Enter` and the indentation is already taken care of or if you want to forgo the `Shift-enter` then you could indeed do a snippet like this.

Sinthoras7:
> Thank you for being honest, but you do understand that I as the reviewer need to understand the code and possibly ask questions about it?

Yes, I understand and because of that it is your right to close this pull request. 
My reasoning behind opening the pull request was that the snippet helped me, so I wanted to share it, as it might help others too. And I thought, maybe the code was understandable to you. 

Now addressing your points:
> ehh, you can't really parse markdown with regex, would be better if you had access to the syntaxtree through the function.

Yes, you are correct that a function with access to the syntax tree would be a better implementation. Nevertheless, I think this implementation works in most cases and is therefore better than not having one at all. (Maybe it could be commented out by default with a comment explaining that it might not work in all cases?)

> Like why priority 50 and 100 was used, that seems excessive.

I think you are correct and with no priority on the regular dm and priority 1 on the new one it still works as expected.

> Besides displaymath in readingview/ markdown spec needs to start on a newline if it spans multiple lines so the only either you press Shift-Enter and the indentation is already taken care of or if you want to forgo the Shift-enter then you could indeed do a snippet like this.

You are correct. That was my fault; I didn't like the large gap between the text and the math in previewmode and not having the newline decreased the gap, but I see that the current behavior has problems and therefore reintroduced the newline.

> ^([ \t]*)(\d+[.)]|[-*+])(\s+)(.*?)\bdm\b$/m
multiline is disabled by latex suite.
it allows 1) but only 1. is markdown syntax and \s+ allows newlines which is not needed anymore in 1.11 where the same indentation level is kept.

I removed the `)` such that only `1.` is allowed now. 
AI rewrote this to not rely on /m and \s+ anymore.

Sinthoras7:
Since I already opened a pr, I added some more snippets that I found useful. For them, I can also answer questions better

superle3:
> I removed the ) such that only 1. is allowed now.

`)` is allowed, it just gets translated to `1.` or a regular html list (which I first wrote that but then used strikethrough on that text)
from the markdown specifications of https://spec.commonmark.org/0.31.2/#lists and https://spec.commonmark.org/0.31.2/#example-307 and more importantly https://spec.commonmark.org/0.31.2/#motivation
> 1. “List markers typically start at the left margin, but may be indented by up to three spaces. List markers must be followed by one or more spaces or a tab.”
> 2. “To make lists look nice, you can wrap items with hanging indents…. But if you don’t want to, you don’t have to.”

you don't have to indent it with 3 spaces but you could?
that would either result in 
```ts
const snippet1 = {
		trigger:
			/(?<=^|\n)(?<firstline>(?<indent>[ \t]*)(?<marker>(?:\d+[.)]|[-+])[ \t]+)(?<text>.*)\b)dm/,
		// three spaces to indicate continuation of paragraph
		replacement: "[[0]]\n   $$\n\t$0\n   $$",
		options: "rtA",
		priority: 1,
	}
// or just in general
const snippet2 = {
		trigger:
			/(?<=^|\n)(?<text>.+\b)dm/,
		// no three spaces for general case but its also valid for lists
		replacement: "[[0]]\n$$\n\t$0\n$$",
		options: "rtA",
		priority: 1,
	}
```

I prefer the second one as obsidian doesn't insert 3 spaces when you press `Shift-Enter`.

> And I thought, maybe the code was understandable to you.

no, it wasn't. I can read it but the intent behind it is lacking for sure the weird greedy modifiers as `*+` and `*?` don't seem to make a lot of sense or the `$` and the stuff after `[ \t]*` after dm. the line on which the regex is triggered is from the beginning of the document till the cursor pos and the regex is surrounded with `(?:<user-regex)$`.

src/default_snippet_variables.js
```diff
@@ -2,4 +2,9 @@
 	"${GREEK}": "(?:alpha|beta|gamma|Gamma|delta|Delta|epsilon|varepsilon|zeta|eta|theta|vartheta|Theta|iota|kappa|lambda|Lambda|mu|nu|xi|omicron|pi|rho|varrho|sigma|Sigma|tau|upsilon|Upsilon|phi|varphi|Phi|chi|psi|omega|Omega)",
 	"${SYMBOL}": "(?:parallel|perp|partial|nabla|hbar|ell|infty|oplus|ominus|otimes|oslash|square|star|dagger|vee|wedge|subseteq|subset|supseteq|supset|emptyset|exists|nexists|forall|implies|impliedby|iff|setminus|neg|lor|land|bigcup|bigcap|cdot|times|simeq|approx)",
 	"${MORE_SYMBOLS}": "(?:leq|geq|neq|gg|ll|equiv|sim|propto|rightarrow|leftarrow|Rightarrow|Leftarrow|leftrightarrow|to|mapsto|cap|cup|in|sum|prod|exp|ln|log|det|dots|vdots|ddots|pm|mp|int|iint|iiint|oint)"
+    "${ACCENT}": "dot|ddot|hat|bar|tilde|vec|underline|overline|mathbf|mathcal|mathrm|mathbb",
+	// ACCENT2 is intentionally identical to ACCENT.
```
superle3: give an example, as that bug should have been solved?
for
```js
const snippet = {
	  trigger: "\\\\(${ACCENT})\\{\\\\(${ACCENT})\\{((?:\\\\(?:${GREEK})|[A-Za-z]))\\}\\}_\\{(\\d+)\\}(\\d)",
	  replacement: "\\[[0]]{\\[[1]]{[[2]]}}_{[[3]][[4]]}",
	  options: "rmA",
	  priority: -1,
	}
```
I get the expansion of 
```
Obsidian Latex Suite: Description: no description provided
Parsed trigger: /(?:\\(dot|ddot|hat|bar|tilde|vec|underline|overline|mathbf|mathcal|mathrm|mathbb)\{\\(dot|ddot|hat|bar|tilde|vec|underline|overline|mathbf|mathcal|mathrm|mathbb)\{((?:\\(?:(?:alpha|beta|gamma|Gamma|delta|Delta|epsilon|varepsilon|zeta|eta|theta|vartheta|Theta|iota|kappa|lambda|Lambda|mu|nu|xi|omicron|pi|rho|varrho|sigma|Sigma|tau|upsilon|Upsilon|phi|varphi|Phi|chi|psi|omega|Omega))|[A-Za-z]))\}\}_\{(\d+)\}(\d))$/
Replacement: \dot{\vec{a}}_{34}
Auto-enlarge brackets: false
```
which is correct?

Sinthoras7: When the snippet
```
{
  trigger: "\\\\(${ACCENT})\\{\\\\(${ACCENT})\\{((?:\\\\(?:${GREEK})|[A-Za-z]))\\}\\}(\\d)",
  replacement: "\\[[0]]{\\[[1]]{[[2]]}}_{[[3]]}",
  options: "rmA",
  description: "Auto subscript after nested modifiers",
  priority: -1,
},
```
is loaded and I type `\dot{\vec{a}}3` nothing happens.

When I add the 2 in the snippet trigger, so the snippet 
```
{
  trigger: "\\\\(${ACCENT})\\{\\\\(${ACCENT2})\\{((?:\\\\(?:${GREEK})|[A-Za-z]))\\}\\}(\\d)",
  replacement: "\\[[0]]{\\[[1]]{[[2]]}}_{[[3]]}",
  options: "rmA",
  description: "Auto subscript after nested modifiers",
  priority: -1,
},
```
is loaded, and then type `\dot{\vec{a}}3`, it gets correctly expanded too `\dot{\vec{a}}_{3}`.

Sinthoras7:
> ) is allowed, it just gets translated to 1. or a regular html list (which I first wrote that but then used strikethrough on that text)
Yes you are correct, I did not know that.

Regarding the indentation:

My understanding of the specs is that you can, but don't have to indent the first level of a list with up to 3 spaces. So a list starts with up to three spaces and then a list marker. 

Furthermore there is  https://spec.commonmark.org/0.31.2/#list-items

> Basic case. If a sequence of lines Ls constitute a sequence of blocks Bs starting with a character other than a space or tab, and M is a list marker of width W followed by 1 ≤ N ≤ 4 spaces of indentation, then the result of prepending M and the following spaces to the first line of Ls, and indenting subsequent lines of Ls by W + N spaces, is a list item with Bs as its contents. The type of the list item (bullet or ordered) is determined by the type of its list marker. If the list item is ordered, then it is also assigned a start number, based on the ordered list marker.

That I understand as follows: For a paragraph to belong to that item it then needs to be indented basically the same amount of characters, that the text succeeding the list marker is.

And if the Paragraph itself starts again with a list marker, then it starts a new sublist.

My proposal for a snippet that does what most users would want in most cases would be: 
1. Go to the beginning of the line and count all spaces and tabs to the first non-space-character. The number of spaces is A_S the number of tabs A_T
2. If there is one there count the width of this list marker W
3. Count the spaces and tabs after the list marker before the first non space. B_S, B_T.

We call (A_S+W+B_S)Spaces+(A_T+B_T)tabs=𝛴.
Then after pressing `dm` insert `\n𝛴$$\n𝛴<cursor>\n𝛴$$\n𝛴<cursor-after-tab>`.

in other words: 
1. Copy everything from the beginning of the line to the first non-space character, call that A.
2. If there is one there count the width of this list marker W.
3. Copy everything after the list marker before the first non-space B

And then call A+(WSpaces)+B=𝛴.

This should work in lists of all indentations. This also triggers in cases where there is no list, but the beginning of the line is indented for some other reason, but arguably this is even desired behavior, because if I indent my line for whatever reason chances are, I also want my math to be on the same indentation level.

To make it a bit more compliant to a list, we could limit B to either ≤4 Spaces or a tab.

superle3:
I don't quite understand what you mean? The current indentation is kept by the snippet engine if you insert a newline, so we could add 3 spaces to indicate a paragraph or dont add 3 spaces and thus keeping the indentation the same as the list marker. But we don't have to keep track of the current indentation?

Sinthoras7:
What exactly do you mean by the current indentation is kept by the snippet engine if you insert a newline?

When I comment out the new snippet therefore just have the snippet
```
{trigger: "dm", replacement: "$$\n$0\n$$", options: "tAw"},
```
and type
```
- Test hello
    - test hello dm
```

This expands to 
```
- Test hello
    - test hello $$
<cursor>
$$

```
and not 
```
- Test hello
    - test hello $$
    <cursor>
    $$

```
which I would expect if the current indentation would be kept. So in which circumstances is the current indentation kept?

By the way: You said, displaymath does not get rendered correctly if it does not starts on a newline: Doesn't this also apply to the default `dm`-Snippet too and it should instead be "**\n**$$\n$0\n$$"?


If the indentation really were to be kept, then it would be better – and I think more aligned with the markdown specs, see my previous message – to insert ((width of the list marker)+1) spaces[^1], but three spaces should cover most use cases, because most list markers have width ≤2 and the extra space does not matter.


[1]: Even better ((width of the list marker)+(number of spaces after the list marker))spaces, but most of the time this should just be one, even though markdown alo

superle3:
> which I would expect if the current indentation would be kept. So in which circumstances is the current indentation kept?

Are you testing on this pr or some earlier version of latex suite as those cases should work?

Sinthoras7:
I was testing on an old version. After updating to the most recent one, this is fixed and indeed the indendation is kept. I am sorry for not noticing that.

src/default_snippet_variables.js
```diff
@@ -2,4 +2,5 @@
 	"${GREEK}": "(?:alpha|beta|gamma|Gamma|delta|Delta|epsilon|varepsilon|zeta|eta|theta|vartheta|Theta|iota|kappa|lambda|Lambda|mu|nu|xi|omicron|pi|rho|varrho|sigma|Sigma|tau|upsilon|Upsilon|phi|varphi|Phi|chi|psi|omega|Omega)",
 	"${SYMBOL}": "(?:parallel|perp|partial|nabla|hbar|ell|infty|oplus|ominus|otimes|oslash|square|star|dagger|vee|wedge|subseteq|subset|supseteq|supset|emptyset|exists|nexists|forall|implies|impliedby|iff|setminus|neg|lor|land|bigcup|bigcap|cdot|times|simeq|approx)",
 	"${MORE_SYMBOLS}": "(?:leq|geq|neq|gg|ll|equiv|sim|propto|rightarrow|leftarrow|Rightarrow|Leftarrow|leftrightarrow|to|mapsto|cap|cup|in|sum|prod|exp|ln|log|det|dots|vdots|ddots|pm|mp|int|iint|iiint|oint)"
+    "${ACCENT}": "dot|ddot|hat|bar|tilde|vec|underline|overline|mathbf|mathcal|mathrm|mathbb",
```
superle3:
```suggestion
    "${ACCENT}": "(?:dot|ddot|hat|bar|tilde|vec|underline|overline|mathbf|mathcal|mathrm|mathbb)",
```

superle3 (commented):
> By the way: You said, displaymath does not get rendered correctly if it does not starts on a newline: Doesn't this also apply to the default dm-Snippet too and it should instead be "\n$$\n$0\n$$"?
if it spans multiple lines and there is a non-whitspace char in the current line, yes (ignoring nested inside html, not sure if that would be different but parser doesn't recognize so ignoring it for now)
If there are only whitespace chars, then not as creating a newline is unnecessary and might even break something.

so 
```ts
const whitespace = {trigger: "dm", replacement: "$$\n\t$0\n$$", options: "tAw"};
const non_whitspace = {trigger: /(?<=\S.*)dm/, replacement: "\n$$\n\t$0\n$$", options: "tAw", priority: 1};
```
would be the replacements of that. (but github doesn't allow me to make that suggestion... cause not inside changed range)


wrap the snippet variables inside non-capturing group instead of inside the snippets, removes clutter.
and the `dm` snippet for lists should approximately follow the spec.

superle3:
Tabstops have priority over matrix shortcuts and tabout thus you cant put a tabstop at the end. Also not sure what the last commit is, as that introduces a syntax error and nothing else.

Sinthoras7:
About the last commit you are correct, that was me in a tired state, we can revert it. 

I understand the problem with the tabstop, but without the extra `\n${indent}$1`, a tab just indents the current line and I do not leave the dm, so there is no tabout. But maybe this is a bug in tabout and should not be taken care of by the snippet?

Also: Is there a specific reason for the tab before `$0` in all dm snippets?

superle3:
> I understand the problem with the tabstop, but without the extra \n${indent}$1, a tab just indents the current line and I do not leave the dm, so there is no tabout. But maybe this is a bug in tabout and should not be taken care of by the snippet?

Wdym exactly? Removing the tabstop works fine for me. Tabout currently only works when you are at the end of the equation (with possible whitespace) for display math (behaviour could be toggled, was waiting for someone to make a pr as they stated they wanted to, but it has been some time so I can also implement it when I have time).

>Also: Is there a specific reason for the tab before $0 in all dm snippets?

That's how you would format it? Like in a latex file, latexindent and tex-fmt indent the part inside the display math delims. Because you are creating a new environment. Doesn't change the rendering.

superle3 (commented):
it now indents the next line if it only contains whitespace or there is no next line. Can I merge it now?
will fix tabout further in #514

Sinthoras7:
Yes you can merge it now. Thank you very much for being so patient with me!

Edit: Does the fix for tabout fix that, when you have a displayed math equation and then tabout of it, the cursor goes to the beginning of the next line and not to the correct indented level?

superle3:
You can test it yourself? It does in the basic case of the next line being empty/not there.

Sinthoras7:
Sorry, I can not test myself, as I did not yet managed to test on the build of the pr. (Before I tested it, with putting the snippets on an obsidian instance, but now after tabout.ts was changed I am unable to test it.)

superle3:
Just pull the changes, build the plugin with `npm run dev` assuming the repo is inside `.obsidian/plugins` and reload your obsidian instance? Suggest you test it in a vault made for testing/sandbox/new vault. I can also merge it and you can give feedback/issue if any later when it's released.

Sinthoras7:
Okay, testing works.

It seems to me that tabbing out now puts the cursor to the correct place, but it does delete all the space before the mathematics. 

Starting situation:
```
1. a
2. b
	1. c <cursor>
```
Now Press `dm` then `d` it works good.
```
1. a
2. b
	1. c 
	   $$
	   	d<cursor>
	   $$
```
Now after pressing tab the situation is the following:
```
1. a
2. b
	1. c 
	   $$
d
	   $$
	   <cursor>
```
So the d is put to the left. This also destroys the markdown formatting, because now the equation does not belong to the point 2.1 again.

src/features/tabout.ts
superle3: Oh yeah needs to be `trimEnd`, made a little sense before as indentation wasn't supported but now not, i'll fix it and then merge it.

Sinthoras7:
Great, thank you very much! :-)

PR #542, opened 2026-04-21
feat: expose API & types for external plugins to suspend features

### LINKED ISSUE #541: [FEATURE] Improve compatibility with the Typst Mate plugin
# Description of the Problem

I would like to improve compatibility so that this plugin can coexist with my plugin, [Typst Mate](https://github.com/azyarashi/obsidian-typst-mate/tree/dev). I often receive requests from users who want to use both plugins at the same time, but at the moment Typst Mate recommends using Typst only inside codeblocks, and users also need to adjust several settings.

# Description of the Solution

I expect the changes to be small. I will describe the details in the PR (#542).

# Additional

> Import feature for LaTeX Suite plugin shortcuts  
> I am undecided between choosing a path of coexistence with LaTeX Suite or importing everything into Typst Mate. I am concerned about performance and conflicts.  
>  
> (Latex suite maintainer) If you want coexistence and there is an issue/something needs to change in latex suite for coexistence to work conveniently, feel free to open up an issue and I'll try to look at. (Adding support for typst shouldn't be that hard I think).  
>  
> For performance just measure/profile and/or check on low battery/quiet mode on a large doc to see if you're disatisfied with it. For latex suite its currently at an alright position? ~~Maybe should look into refusing tha math rendering that obsidian uses instead of rendering the same eq? But that's messing with internal APIs which break easier.~~ nvm more like disabling it when I insert a key press into the history and immediately undo it right after. Obsidian doesn't render inline if it's revealed  
>  
> And for conflicts, as long as the user doesn't turn on conflicting features it should be alright (like turning on 2 snippet engines).  

_Originally posted by @superle3 in [#29](https://github.com/azyarashi/obsidian-typst-mate/issues/29#issuecomment-4283407415)_

azyarashi:
Fix: #541

### ~~What I want to do~~

~~When the cursor is inside a math that is rendered by Typst~~

- ~~Disable Obsidian LaTeX Suite's math-context detection ([`src/utils/context.ts`](https://github.com/artisticat1/obsidian-latex-suite/blob/main/src/utils/context.ts)).~~
- ~~Add a setting to disable snippets when the snippet `flags` contains `m`, `M`, or `n`.~~

### ~~Background~~

~~To make this possible, starting from the next version, Typst Mate will provide an API that is registered as a global variable (`window`/`globalThis`). It is planned to include properties such as `isInsideTypst: boolean` and `mathFlag: "m" | "M" | "n"`.~~

### ~~Notes~~
~~If there is anything else that is needed, please let me know.~~

~~These changes will be done after Typst Mate v3.0.0 is ready and after its TypeScript type definitions become available to reference, so this PR will stay as Draft for a while.~~


superle3:
> - Disable Obsidian LaTeX Suite's math-context detection ([src/utils/context.ts](https://github.com/artisticat1/obsidian-latex-suite/blob/main/src/utils/context.ts)).


(You mean `options` cause flags means regex flags and those are not allowed/illegal?)
Thought another viewplugin would be enough that wraps around tex-mate's api and point https://github.com/artisticat1/obsidian-latex-suite/blob/main/src/utils/context.ts#L240 towards the typst viewplugin when typst toggle is on.

> - Add a setting to disable snippets when the snippet flags contains m, M, or n.

why does a toggle to disable math snippets need to be added? one could add a flag in their snippets `includeMath` and filter the snippets `[...].filter(s => !/[nmM]/.test(s.options)`

> To make this possible, starting from the next version, Typst Mate will provide an API that is registered as a global variable (window/globalThis). It is planned to include properties such as isInsideTypst: boolean and mathFlag: "m" | "M" | "n".

Does typst have a mode thats semantically neither inline math or display math but still counts as math? otherwise "M" | "n" would be enough  (and "inline" | "display" would be a bit more readable, "n" | "M" is used in snippets cause those are more shorthand but here shouldn't be needed in the source code).

(if this was just an outline then its fine for now and you can partly ignore my comments if they don't make sense down the line.)

azyarashi:
Sorry, I didn't explain enough 😭

> Disable Obsidian LaTeX Suite's math-context detection (`src/utils/context.ts`).

I haven't been able to fully read the code yet, but features like Conceal, Tabout, and Color and highlight matching brackets can't work without context, right?

Since Typst syntax and TeX syntax are different, when the cursor is inside a Typst math region, I think those features should be handled by Typst Mate, rather than being interpreted as TeX math by this plugin.

(Note: Typst Mate does not replace all math with Typst. Depending on a user-configured prefix or YAML frontmatter, it can switch between MathJax and Typst. So I'll need to expose an API from Typst Mate.)

---

> why does a toggle to disable math snippets need to be added?

This is for the case where:

- the snippet flags contain `m`, `M`, or `n`,
**and**
- the cursor is inside math that is rendered by Typst.

In that situation, we can prevent MathJax-oriented snippets from expanding while the user is actually editing Typst math.

Typst Mate will provide Typst-specific snippets (Typst needs to distinguish between markup mode, math mode, and code mode), so I was thinking it would be better to separate responsibilities between the snippet sets.

---

> Does typst have a mode thats semantically neither inline math or display math but still counts as math? otherwise "M" | "n" would be enough (and "inline" | "display" would be a bit more readable, "n" | "M" is used in snippets cause those are more shorthand but here shouldn't be needed in the source code).

Since it can be represented as `inline` | `display`, I'll go with that.

superle3:
> I haven't been able to fully read the code yet, but features like Conceal, Tabout, and Color and highlight matching brackets can't work without context, right?

Yes?  But they just need the mathbounds and like a wrapper to check if the main cursor is in a math bound.

Tabout would need different delimiter maps though conceal probably wouldn't work as typst is more complicated than mathjax that a simple regexp wouldn't work.

azyarashi:
In other words, when the cursor is inside Typst, I will intentionally disable the context and turn those features off. I had planned to provide that determination by my api.

Although the implementation is quite different under the hood, features such as Conceal, Tabout, and Color and highlight matching brackets for Typst already exist in my plugin. Right now, there is a setting in my plugin to turn them on and off so that they do not conflict with LaTeX Suite.

superle3:
So only the viewplugins in `context.ts` need to be public in latex suite/wrapper `disableMathMode` so you can turn them off with another viewplugin? (With priority high or lower).

azyarashi:
Do you mean that I should handle it by manipulating the viewplugin? I'll give it a try.
Thank you!

azyarashi:
Thanks for implementing this!
Just to clarify, is it correct that I should keep calling disableMath whenever I want to disable the LaTeX Suite features?

superle3:
Yeah something like

```ts

class Test implements PluginValue {

  update(update) {
   const isTypst= ...
   try {
     app.plugins.getPlugin("obsidian-latex-suite")?.disableMath(update.view)
   } catch {}
  }
}

```

And it won't recompute its values
 (not the same as disabling features, as it doesn't recompute its values when document hasnt changed and/or the selection hasn't changed but it should cover your usecase).

azyarashi:
I confirmed it works properly! Thank you so much!

azyarashi:
Would it be okay to add a type definition file? If so, where should I put it? For now, I placed it at the repository root as `api.d.ts`.

(I referred to Obsidian Excalidraw. Obsidian Excalidraw puts its type definition files in the `docs/API` directory. Also, it might be worth considering switching to Rolldown in the future.)

cf68a1d72e4ca286d875916da9933d5a2e2d2613

superle3:
> Would it be okay to add a type definition file? If so, where should I put it? For now, I placed it at the repository root as api.d.ts.

For now in root of `src` is enough as I won't be generating the type files. 
Instead I'll make sure the plugin satisfies the public api with `implements api`.

Maybe in the future I'll look into how to do things with `@public` tags such that other methods don't get included, but for now it would too complex for a small api.

azyarashi:
Got it. Please remove `rollup`, `rollup-plugin-dts` and `build:types` from `package.json`🙏

PR #557, opened 2026-05-30
wip: add folder to snippets.

### LINKED ISSUE #58: [FR] Add an option to specify the folders in which a snippets functions
I think it's good to have the ability to restrict a snippet to only expand in certain folders . This is useful in use cases like having several folders for different subjects (math, physics etc), thus need different set of snippets .

One possible approach to implememt this is to add a (optional) field in each snippet's definition like
```
path: [ "path/one", "path/two" ] 
```
Besides it will be even better if we'll have some kind of glob support.

Thanks for you great work! : )

superle3:
Not sure about how to get the path of the file into the run_snippets function.

### DIFF
--- a/esbuild.config.mts
+++ b/esbuild.config.mts
@@ -88,7 +88,10 @@ if (dev) {
 	}, 5000);
 	fs.watch("src/parser", { recursive: true }, debouncedExec);	
 } else if (prod) {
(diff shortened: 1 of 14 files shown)


superle3:
Fixes #58

PR #567, opened 2026-06-03
[FEATURE] Custom conceal characters

### LINKED ISSUE #242: [FEATURE] Custom conceal characters
# Description of the Problem
There are some sequences where either there is no current concealment or I can think of a better but I can't just make it do that. For example, I would like to have `\lVert a \rVert` be concealed as `|| a ||` but it currently just shows the base LaTeX.

# Description of the Solution
Add JSON like with the snippets mapping commands or strings of text to a concealment. For example,
```json
```

# Additional

superle3:
Fixes #242

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -522,6 +522,48 @@ When 2 or more keymaps have the same key, the following order will be used
 
 If the feature is disabled, the keymap will be skipped.
 
+## Conceal maps
+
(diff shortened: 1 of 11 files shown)


superle3:
hmmm, maybe an api would be better? don't like that there are now 3 files/folders to be watched.

PR #568, opened 2026-06-03
wip basic luasnip api internal only for now

### LINKED ISSUE #558: [FEATURE] luasnip api
# Description of the Problem

Currently for visual snippets or full equation snippets, its impossible to escape the tabstop syntax and the capture syntax.

# Description of the Solution

Without breaking backwards compatibility, add a simple api such as `ls.tabstop`, `ls.captures`, `ls.text`.

# Additional
alternative is breaking comparability or an extra version. Related to #49 and #282.

superle3:


superle3 (commented):
only exponsing `text_node`, `tabstop_node` and `capture_node` as the `snippetNode` from luasnip requires some refactoring to allow nesting tabstops and that becomes a bit too much for this pr.

PR #572, opened 2026-06-14
New snippet; "deg" for degree symbol

skelobones:
Added snippet to turn "deg" into a degree symbol. Line is placed at the start of trigonometry for easier disabling if wanted. Set apart from sine, cos & tan snippets.


superle3 (commented):
I think symbols make more sense in this case and since for the greek symbols and all the other symbols, macros are used instead of unicode, it will be kept as a macro (feel free to change it in your own snippets though).
Thank you for the contribution

PR #574, opened 2026-06-27
feat:  saving snippets in hidden folders

### LINKED ISSUE #67: [FR] Saving Snippets in a Hidden Folder
It would be great if the plugin can read a snippets file in a hidden directory like `.config/`

superle3:
Fixes #67

allow files to be saved in hidden folders or outside the vault on desktop.


superle3 (commented):
hopefully this doesn't cause any issues, but its only reading so shouldn't be any negative affects as we are only reading

PR #575, opened 2026-06-28
Superle3/issue122

### LINKED ISSUE #122: [Bug] math/text enviroment is not determined correctly with unusual tokens
Triggering a snippet like `{trigger: "h", replacement: "\\colorbox{$0}{\\\(${VISUAL}\\\)}$1", options: "mA"}` often causes `\(`, the alt math mode token be converted to `\\left(`, which renders the expression invalid as it is disallowed in text mode.
`$\colorbox{}{\\left( \frac{1}{2}\ \right)}$`

Also, latex suite treats `$\text{\(           \)}$` as a text environment. Please add support for all environment tokens, or a setting to add them.

superle3:
Fixes #122

doesn't fix the case for snippets expanding in colorbox but does fix the case of wrongly auto enlarging math delimiters

### DIFF
--- a/src/editor_extensions/highlight_brackets.ts
+++ b/src/editor_extensions/highlight_brackets.ts
@@ -31,6 +31,7 @@ const bracket_delimiters = [
 	["{","}"],
 	["[","]"],
 	["(",")"],
+	["\\(", "\\)"],
 	// these don't neccessarily have to be paired, but they often do so we treat them like a pair.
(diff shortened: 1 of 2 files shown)


superle3 (commented):
`\[\]` doesn't work in mathjax and `\(\)` should be one bracket type (if its inside text but todo for later when the parser is more advanced).

Issue #250, opened 2024-01-31
[FEATURE] Can CANCEL support rendering textcolor{}{} ?


Jeffliv5:
# Description of the Problem

I write some snippets to quickly annotate Latex using textcolor{}{}. While the CANCEL cannot render Latex with textcolor{}{} correctly. You can look at the below picture to understand what I am saying.
![image](https://github.com/artisticat1/obsidian-latex-suite/assets/67423287/48acb64b-c636-4b57-a911-6263e31d9282)

# Description of the Solution

# Additional

I would like to share with you my snippets to quickly annotate the Latex:
` {trigger: "R", replacement: "\\textcolor{red}{ ${VISUAL} }", options: "mA", description: "change color to red, an easy way to annotate math"}, `
` {trigger: "B", replacement: "\\textcolor{blue}{ ${VISUAL} }", options: "mA", description: "change color to blue, an easy way to annotate math"},`
Latex suite is really helpful, thanks for your hard work.

superle3:
You mean CONCEAL instead of CANCEL?

Issue #207, opened 2023-10-05
[FEATURE] Enable snippets between $$ in code block (for pseudocode)


phantamoe:
I know this might sound weird, but I want the snippets to work when I'm in a code block when it's between $$. Sometimes I need to type [pseudocode](https://github.com/ytliu74/obsidian-pseudocode), where $$ appear quite often. Right now option c only supports regular text in a code block, and I cannot limit the scope of the snippets to function specifically between $$ with mcA.

superle3:
We would need our own parser at that point. Which isn't that bad but could be a bit complicated so low priority for now.

PR #581, opened 2026-07-12
refactor: create and integrate our own parser

### LINKED ISSUE #153: [BUG] Math mode snippets trigger inside \text{}

### LINKED ISSUE #547: [BUG] Shift+Enter exits too far in block matrix environments when closers share a line
# Description

Shift+Enter in matrix environments can exit too far when the current matrix is nested inside another math structure and multiple closing boundaries are on the same line.

For example, when editing a `bmatrix` inside a `\left\{ ... \right\}` set, pressing Shift+Enter at the end of the final matrix entry can move the cursor outside the surrounding `\right\}` instead of stopping immediately after `\end{bmatrix}`.

# To Reproduce

Example 1: `bmatrix` inside `\left\{ ... \right\}`

1. Create a nested math structure like this.
2. Place the cursor at the end of the final matrix entry, marked here with `▶`.
3. Press Shift+Enter.

```latex
\left\{
\begin{bmatrix}
-2 \\
1 \\
0
\end{bmatrix},
\begin{bmatrix}
1 \\
0 \\
1 ▶
\end{bmatrix} \right\}
```

Actual behavior:

```latex
\left\{
\begin{bmatrix}
-2 \\
1 \\
0
\end{bmatrix},
\begin{bmatrix}
1 \\
0 \\
1
\end{bmatrix} \right\} ▶
```

Example 2: `bmatrix` inside `align`

```latex
\begin{align}
\begin{bmatrix}
1 & 2 & 3 ▶
\end{bmatrix} \end{align}
```

Actual behavior:

```latex
\begin{align}
\begin{bmatrix}
1 & 2 & 3
\end{bmatrix} \end{align} ▶
```

# Expected Behavior

Shift+Enter should exit only the current innermost matrix-like environment, rather than moving to the end of the next physical line.

For the `\left\{ ... \right\}` example, I would expect:

```latex
\left\{
\begin{bmatrix}
-2 \\
1 \\
0
\end{bmatrix},
\begin{bmatrix}
1 \\
0 \\
1
\end{bmatrix} ▶ \right\}
```

For the `align` example, I would expect:

```latex
\begin{align}
\begin{bmatrix}
1 & 2 & 3
\end{bmatrix} ▶ \end{align}
```

Conceptually, when the cursor is inside a configured matrix environment such as `bmatrix`, `pmatrix`, `matrix`, etc., Shift+Enter should move the cursor to just after that environment’s closing symbol:

```latex
\begin{bmatrix}
1 & 2 & 3
\end{bmatrix} ▶
```

It should not also exit surrounding structures such as `\right\}` or `\end{align}`.

# Additional Info

The relevant implementation appears to be in `src/features/matrix_shortcuts.ts`:

```ts
const taboutMatrixShortcutCallback = (view: EditorView, bounds: Bounds): boolean => {
    const ctx = getContextPlugin(view);
    if (ctx.mode.blockMath) {
        // Move cursor to end of next line
        const d = view.state.doc;

        const nextLineNo = d.lineAt(ctx.pos).number + 1;
        const nextLine = d.line(nextLineNo);

        setCursor(view, nextLine.to);
    }
    else if (ctx.mode.inlineMath) {
        setCursor(view, bounds.outer_end);
    }
    return true;
}
```

So in block math, the shortcut moves to `nextLine.to`, while in inline math it uses `bounds.outer_end`.

Since `matrixShortcutsRunner(...)` already computes `envBounds` using `ctx.isWithinEnvironment(...)`, I wondered whether the block-math branch could also use the detected environment boundary instead of the end of the next line. For example, maybe this could be handled with `bounds.outer_end`, similar to the inline-math branch. That said, I do not want to assume the fix is that simple. There may be a reason the block-math branch currently uses `nextLine.to`.

I realize Latex Suite is not a full LaTeX parser, so this may be more of an edge-case bug / implementation limitation. The current behavior appears intentional from the comment “Move cursor to end of next line,” but it produces surprising behavior in nested math structures.

Related history I found while checking for duplicates:

- #529 
- #421
- #394

#529 looks related, but from what I can tell it addressed the inline-math Shift+Enter path, while the current block-math branch still moves to the end of the next physical line.

### LINKED ISSUE #580: [BUG] Wrong pairing in multiple brackets
# Description
Wrong automatic large pairing in multiple brackets

# To Reproduce

Image

superle3:
switch from integrated hypermd parser
to our own incremental parser with lezer/markdown and texlyre's latex parser.

This makes the concealment and highlighting faster and hopefully offsets the cost of the incremental parser.

Fixes #580 
Fixes #547
Fixes #153
hopefully addresses #320,

would allow #207 and #250 to be addressed.


superle3:
larger pr than I would like but parser implementation was depended on how it was used in the concealment and such so was doing both at the same time and now dont wanna spend time to extract them.

PR #605, opened 2026-08-17
fix: hypermd lang detector

### LINKED ISSUE #603: [BUG] Note source view is treated as math mode
# Description
Whenever I type in source mode, the whole document is treated as if it’s in math mode. This means that I can’t write normal text without triggering the default math snippets. Even the preview shows at the top of the document.

# To Reproduce
I tried reinstalling the plugin from scratch and it still behaves like this.

# Expected Behavior
Math mode snippets should not be triggered in normal text mode. The preview should only show in math mode.

# Additional Info

Image

superle3:
It seems in some cases its not detected correctly so assuming excalidraw's content stays the same instead. Like with templater that overrides the parser and sadly also the language's name.

Fixes #603

### DIFF
--- a/src/parser/language.ts
+++ b/src/parser/language.ts
@@ -4,13 +4,14 @@
  * MIT License
  * Copyright (C) 2018-2021 by Marijn Haverbeke <marijnh@gmail.com> and others
  */
-import { language, ParseContext } from "@codemirror/language";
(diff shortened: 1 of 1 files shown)


superle3:
github is down so can't merge.... whyy

PR #613, opened 2026-08-19
fix: move color and colorbox macros to textArea (fixes #612)

### LINKED ISSUE #612: [FEATURE] Allow text-mode snippets inside color and colorbox macros
# Description of the Problem
Commit `37ec346` moved `color`, `textcolor`, `colorbox` and `fcolorbox` into `snippetLessArea`, which sets their environment to `"none"` and completely disables all snippets. 
This breaks intentional text-mode (`tA`) workflows used for quickly inserting color names, such as:
- `{ trigger: "li", replacement: "lime", options: "tA" }`
- `{ trigger: "cy", replacement: "cyan", options: "tA" }`

# Description of the Solution
Move these macros back from `snippetLessArea` into the `textArea` array, ensuring `textcolor` retains its bounds via `{ name: "textcolor", arguments: [0] }`. 

# Additional
- **Alternatives considered:** Using prefix triggers (e.g., `{ trigger: "cli", replacement: "\\color{lime}", options: "tA" }`), but this disrupts editing workflows when modifying existing equations.
  
  **Why this is the right approach:**
  - **Preserves safety:** Placing them in `textArea` evaluates the environment as `"text"`, which natively blocks math-only (`m`) snippets and prevents the accidental expansion bugs.
  - **Restores functionality:** It successfully enables text-mode (`t`) snippets to expand properly for text-literal arguments like colors.

- **Context:** Since these commands take plain text/literals as arguments, `textArea` is the semantically correct place for them.

Remyanscar:
### Description
Resolves #612

This PR moves `color`, `textcolor`, `colorbox` and `fcolorbox` from `snippetLessArea` back to `textArea`. 

- **Preserves safety:** Evaluates the environment as `"text"`, natively blocking math-only (`m`) snippets to prevent rendering bugs.
- **Restores functionality:** Allows text-mode (`tA`) snippets to expand properly for text-literal arguments (e.g., `li` -> `lime`).

### Testing
- `\color{li}` successfully expands to `\color{lime}`.
- Math snippets remain blocked inside color/colorbox arguments.

### DIFF
--- a/src/utils/default_text_areas.ts
+++ b/src/utils/default_text_areas.ts
@@ -27,6 +27,10 @@ export const textArea = [
 	{ name: "mbox" },
 	{ name: "fbox" },
 	{ name: "framebox" },
+	{ name: "textcolor", arguments: [0] }, // only the first argument is text/color, the second argument is math
+	{ name: "color" },
+	{ name: "colorbox" },
+	{ name: "fcolorbox" }, // has two inputs \fcolorbox{color}{background}{text} needs seperate handling
 ] as const satisfies readonly MacroArea[];
 
 /**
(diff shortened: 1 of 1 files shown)


superle3:
I won't do it this way, but I'll add support for it. Also the example isn't that good as li would be used for light in normal text or.

PR #631, opened 2026-09-06
snippet: optimize conflicts

### LINKED ISSUE #68: [Feature Request] Disable snippets when writing Latex commands
If the cursor is positioned as “\blahblahbl<cursor>”, disable the snippet completion. This could be done with a regex that checks that there is a backslash followed by text not separated by spaces.

The reason for this is to be able to write things manually for less-often used symbols. For example, \land triggers the “and” snippet for me, so I write the “d” first and then the rest before it to avoid triggering the snippet. I know that this can be fixed by customizing the snippets shortcuts, yet I think this would still be a powerful feature.

superle3:
optimize to remove conflicts like \leq1 to generalize it towards any macro. Also easier writing for automatic and triggerKey snippets.

related #630


superle3 (commented):
Fixes #68 and all its duplicates.
This was already possible but now the default snippets should avoid them as much as possible.

PR #632, opened 2026-09-06
fix: treat begin as command

superle3:
`\begin` and `\end` have a different name/argument structure in the parser causing it not to be recognized as a macro and thus is treated as math while it should be snippetless.


superle3 (commented):
also added the test back that test the snippets context, somehow added it locally but then deleted it or its still on some local branch

PR #636, opened 2026-09-07
fix: selection tabstop highlight

superle3:
Because EditorView.decorations was used instead of EditorView.outerDecorations, the decorations were split up and the outlines overlap causing a slight visual disturbance.


superle3:
# Before

image

where the selections overlaps with `\times` because the syntax highlighter, highlights them.

# after

image

PR #638, opened 2026-09-07
fix: keep checking excludedMacros through nested environments

### LINKED ISSUE #629: [FEATURE] Let excludedMacros apply inside environments nested in the excluded macro
# Description of the Problem

`excludedMacros` stops applying as soon as there is an environment between the cursor and the excluded macro. In `isWithinExcludedScope` (src/snippets/snippets.ts), the stack walk returns `false` on the first environment it meets, so the macro further out is never checked:

```ts
} else if (envName.kind === "environment") {
    return false;
} else if (isMacroArgumentCount(envName, this.excludedMacros)) {
    return true;
}
```

Minimal reproduction with the default `->` snippet, which has `excludedMacros: ["ce"]`:

```latex
$$ \ce{ A -> B } $$                                   % -> is left alone, as expected
$$ \ce{ \begin{align} &A -> B \\ &C -> D \end{align} } $$   % -> expands to \to
```

The same happens to the auto-subscript snippet (`excludedMacros: ["pu", "ce"]`): `H2O` becomes `H_{2}O` inside `\ce{ \begin{align} ... }`. mhchem accepts `align` inside `\ce{}`, and multi-line redox half-reactions are a natural place to use it, so the exclusion in a pretty common chemistry use case.

`excludedEnvironments` cannot express this. Excluding `->` from `align` would disable it in every `align`, not only those inside `\ce{}`.

## Why this is not the same as the mode rule

The comment next to that early return explains that environments always override the scope while macros can inherit it, with `\textcolor` inside `\text` as the example. That rule is about **mode** (math vs text), and I am not proposing to change it. `\text{ \begin{align} ... }` should still be math inside the environment.

`excludedMacros` answers a different question, about **syntax scope**: is the cursor inside a macro whose argument has its own grammar (`\ce`, `\pu`, `\bond`, `\label`)? An environment inside `\ce{}` is still mhchem syntax. Today both questions share one walk, so the mode rule accidentally governs the scope check too.

# Description of the Solution

Keep the environment short-circuit for mode decisions, but let the `excludedMacros` walk continue outward through environments until it reaches the math boundary:

```ts
isWithinExcludedScope(stack: StackOutput[]): boolean {
    if (this.excludedEnvironments.length === 0 && this.excludedMacros.length === 0) return false;
    for (const envName of stack) {
        if (envName.kind === "math") return false;
        if (envName.kind === "environment") {
            if (this.excludedEnvironments.includes(envName.name)) return true;
            continue; // an environment does not end the enclosing macro's syntax scope
        }
        if (isMacroArgumentCount(envName, this.excludedMacros)) return true;
    }
    return false;
}
```

Scope of the change:

- Only snippets that already declare `excludedMacros` are affected, and only when an environment sits between the cursor and that macro. Nothing changes for snippets without exclusions, for `excludedEnvironments`, or for text/math mode detection.
- An exclusion is already an explicit per-snippet opt-in by the user, so I think tunnelling through environments should be the default rather than a separate flag. If you would rather keep the current behaviour as the default, a per-snippet option would also solve my case.
- `isWithinIncludedScope` (unreleased `includedMacros`) looks only at `stack[0]`, so it has the same limitation. Happy to cover it in the same PR if you agree with the direction.

I can open a PR for this if the approach sounds right.

Related: #248 (where `excludedMacros` was requested), #556 (implementation), #577 (refactor).

Plugin version: 1.12.8.

Comprehensive-Jason:
Fixes #629

`isWithinExcludedScope` returned early on any environment that was not itself excluded, so a snippet's `excludedMacros` never applied once an environment sat between the cursor and the macro, e.g. `\ce{ \begin{align} A -> B \end{align} }` still expanded `->`. As discussed in #629, environments are now skipped instead, which matches what `Context.isWithinMacros` already does.

Changes in `src/snippets/snippets.ts`:

- `isWithinExcludedScope`: a non-excluded environment `continue`s the walk instead of returning `false`; `excludedEnvironments` and the math boundary behave as before.
- `isWithinIncludedScope`: skips environments too, but still only considers the innermost macro, so an included macro further out does not re-enable snippets.

`npm run lint` and `npm run build` pass. No test added, since the existing snippet tests run through the Obsidian integration harness; I will report a manual check with the default `->` and auto-subscript snippets inside `\ce{ \begin{align} ... }` on the issue.

### DIFF
--- a/src/features/run_snippets.ts
+++ b/src/features/run_snippets.ts
@@ -90,7 +90,7 @@ const runSnippetCursor = (view: EditorView, ctx: Context, snippetInfo: SnippetIn
 	if (snippetInfo.key && snippetInfo.key.length !== 1) {
 		return {success: false, shouldAutoEnlargeBrackets: false};
 	}
-	const envNames = Array.from(ctx.getEnvNames())
+	const envNames = Array.from(ctx.getEnvNames(to))
 	const updatedLine = line + key;
 	for (let i=0; i < snippetInfo.snippets.length; i++) {
 		const snippet = snippetInfo.snippets[i];
--- a/src/snippets/snippets.ts
+++ b/src/snippets/snippets.ts
@@ -154,13 +154,12 @@ export abstract class Snippet<T extends SnippetType = SnippetType> {
 	isWithinExcludedScope(stack: StackOutput[]): boolean {
 		if (this.excludedEnvironments.length === 0 && this.excludedMacros.length === 0) return false;
 		for (const envName of stack) {
-			if (envName.kind === "environment" && this.excludedEnvironments.includes(envName.name)) {
-				return true;
+			if (envName.kind === "environment") {
+				if (this.excludedEnvironments.includes(envName.name)) return true;
+				// An environment does not end the enclosing macro's scope: \begin{align} inside \ce{}
+				// is still mhchem syntax, so keep walking outward to check excludedMacros.
+				continue;
 			} else if (envName.kind === "math") {
-				return false;
-				// environments always override the scope whereas macros can sometimes take scope from outer macro
-				// like \textcolor is math or text depending if its inside a \text macro or not.
-			} else if (envName.kind === "environment") {
 				return false;
 			} else if (isMacroArgumentCount(envName, this.excludedMacros)) {
 				return true;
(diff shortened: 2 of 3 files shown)


Comprehensive-Jason:
Manually verified with a build of this branch installed through BRAT (`1.12.8-fix629` on my fork), Obsidian on Windows, Latex Suite 1.12.8 base.

Inside `$$ \ce{ \begin{align} ... \end{align} } $$` with the default snippets:
- `->` (`excludedMacros: ["ce"]`) is no longer expanded to `\to`.
- `H2O` (auto-subscript, `excludedMacros: ["pu", "ce"]`) is no longer turned into `H_{2}O`.

Both still expand as before in plain math and in an `align` that is not inside `\ce{}`, and `excludedEnvironments` behaves unchanged.

superle3 (commented):
also added tests so some refactor later doesn't accidentally break it.

PR #641, opened 2026-09-09
fix: mount settings editors in the Settings window's document

### LINKED ISSUE #640: [BUG] Snippet editors in Settings render unstyled on Obsidian 1.14 (CodeMirror base theme mounts into the wrong document)
# Description of the Problem

Since 1.13.0, the "Snippets" and "Snippet variables" editors in the settings tab render broken on Obsidian 1.14: the text sits below and outside the editor box and overflows the settings underneath it. The box itself is drawn at the right size, only its contents are misplaced.

Obsidian 1.14 renders Settings in a separate window with its own `document`. `createCMEditor` in `src/settings/settings_tab2.ts` constructs the `EditorView` with only `state` and `parent`, so CodeMirror falls back to the main window's `document` as its root and mounts its base theme (the `.ͼ1 { display: flex !important; ... }` style module) there. The Settings window never receives that stylesheet, and without it `.cm-gutters` and `.cm-content` lay out as plain blocks.

Measured in the Settings window (Obsidian 1.14.1, Latex Suite 1.13.0, Windows):

- `.cm-editor` box: 562 x 300 px, computed `display: block`, `position: static`
- `.cm-scroller`: 832 px tall, `overflow: visible`
- `.cm-content` starts 435 px below the top of the editor and is 398 px tall, so it covers the next setting items
- the Settings document has no `<style>` containing `.ͼ1`, while the main document does

# Reproduction

1. Obsidian 1.14.x, Latex Suite 1.13.0, "Load snippets from file" off (or "Load snippet variables from file" off).
2. Settings > Latex Suite > Snippets. The Snippets editor (or, under Advanced snippets, the Snippet variables editor) shows its text outside the box.
3. Here's an image of the issue: 
Image

# Description of the Solution

Pass the element's document as the editor root:

```ts
const view = new EditorView({
    state: EditorState.create({ doc: content, extensions }),
    parent: node,
    root: node.ownerDocument,
});
```

Verified in the live Settings window: constructing an `EditorView` with `root` set to that document mounts the base theme into it, and the editor comes up as `display: flex` / `position: relative` with a scrolling `.cm-scroller`. PR to follow.

Comprehensive-Jason:
Fixes #640

Obsidian 1.14 renders Settings in its own window with its own `document`. `createCMEditor` constructed the `EditorView` with only `state` and `parent`, so CodeMirror defaulted its root to the main window's `document` and mounted the base theme (`.ͼ1 { display: flex !important; ... }`) there. The Settings window never got that stylesheet, so the Snippets and Snippet variables editors rendered as plain blocks with their content spilling out below the box.

Passing `root: node.ownerDocument` alone turned out not to be enough (first commit, tested with a BRAT build): Obsidian builds the settings DOM in the main window and migrates it into the Settings window afterwards, so `ownerDocument` is still the main document at construction time. The second commit keeps that and additionally re-points the root with `EditorView.setRoot` from Obsidian's `onNodeInserted` and `onWindowMigrated` hooks on the editor's parent element, so the styles follow the editor into whichever document it ends up in.

Verified in a live 1.14.1 Settings window that both hooks fire when a detached main-window element is appended into that window, that `view.root` stays the main document until `setRoot` is called, and that after `setRoot` the base theme is mounted there and the editor lays out as `display: flex` with a scrolling `.cm-scroller`. `npm run lint` and `npm run build` pass. On Obsidian versions that render Settings in the main window this is a no-op, since the root already matches.

### DIFF
--- a/src/settings/settings_tab2.ts
+++ b/src/settings/settings_tab2.ts
@@ -814,15 +814,24 @@ class ConfirmationModal extends Modal {
 	}
 }
 
-function createCMEditor(content: string, extensions: Extension[], node: Element) {
+function createCMEditor(content: string, extensions: Extension[], node: HTMLElement) {
 	const view = new EditorView({
 		state: EditorState.create({ doc: content, extensions }),
 		parent: node,
-		// Obsidian 1.14 renders Settings in its own window. Without an explicit root, CodeMirror
-		// mounts its base theme into the main window's document, and the editor in the Settings
-		// window renders unstyled (gutter and content stacked as blocks, overflowing the box).
 		root: node.ownerDocument,
 	});
+
+	// Obsidian 1.14 renders Settings in its own window, and it builds the settings DOM in the
+	// main window before moving it there. CodeMirror mounts its base theme into the root it was
+	// given, so once the editor lands in a different document its styles are missing and the
+	// gutter and content render as plain blocks overflowing the box. Re-point the root whenever
+	// the editor is inserted or migrated into another window.
+	const syncRoot = () => {
+		const doc = view.dom.ownerDocument;
+		if (view.root !== doc) view.setRoot(doc);
+	};
+	node.onNodeInserted(syncRoot, true);
+	node.onWindowMigrated(syncRoot);
 
 	return view;
 }
--- a/src/settings/settings_tab2.ts
+++ b/src/settings/settings_tab2.ts
@@ -818,6 +818,10 @@ function createCMEditor(content: string, extensions: Extension[], node: Element)
 	const view = new EditorView({
 		state: EditorState.create({ doc: content, extensions }),
 		parent: node,
+		// Obsidian 1.14 renders Settings in its own window. Without an explicit root, CodeMirror
+		// mounts its base theme into the main window's document, and the editor in the Settings
+		// window renders unstyled (gutter and content stacked as blocks, overflowing the box).
+		root: node.ownerDocument,
 	});
 
 	return view;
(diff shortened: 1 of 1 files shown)


Comprehensive-Jason:
Pushed a second commit: the `ownerDocument` fix alone did not hold up in a BRAT build, because Obsidian builds the settings DOM in the main window and migrates it into the Settings window afterwards. The editor now re-syncs its root via `setRoot` from `onNodeInserted` / `onWindowMigrated`. PR description updated.

Comprehensive-Jason:
Manually verified the second commit with a BRAT build (`1.13.0-fix640-2` on my fork) on Obsidian 1.14.1, Windows, after a full relaunch so the Settings window started without any stylesheet left over from earlier debugging.

Both editors were affected and both are fixed: with "Load snippets from file or folder" off, the Snippets editor, and with "Load snippet variables from file or folder" off, the Snippet variables editor under Advanced snippets, now render inside their boxes with a scrolling content area, and the settings below them are no longer covered. The first commit alone (`root: node.ownerDocument` only) did not fix either on the same setup, which is what led to the `setRoot` re-sync.

superle3:
its also in 1.13, I don't test with popup settings cause they are annoying to test with and block the command pallete from appearing in the main window.

superle3:
activeDocument should have been used as the root, and it does work without the moving, but I'll keep it in case obsidian does something weird.

hmm, uhhh nvm, just gonna do `activeDocument` alone, that should work.

PR #649, opened 2026-09-15
fix: mobile single keymaps

superle3:
autofraction wasn't being triggered by `/` because the keymaps weren't being run.


superle3 (commented):
oh duplicate of #653

